PR #12, opened 2018-07-30
Indent Fix

prashant-andani:
Indent fix at the loader.js


jxnblk:
Thanks, but this is very hard to see what the "fix" is due to the changes from prettier. I'd rather introduce prettier as a separate PR later if you can undo those changes

jxnblk:
Thanks for the PR but I think I’d rather close this out and set up prettier later when things are a little more settled in the code base

PR #18, opened 2018-07-30
Add ability to ignore key events

johno:


jxnblk:
Thanks! Looking at the code change, I wish I'd gone ahead and made KeyboardShortcuts its own component, but can follow up with that later

PR #21, opened 2018-07-31
Improve slide parsing

johno:
Currently it's not possible to use tables because they're
parsed as slides. This ensures that a slide delimiter starts
at the beginning of the line.


busypeoples:
Not sure if this would work, as we're dealing with a string. The delimiter would only work at the very start of the string, maybe I'm missing something here.
Wouldn't using `/\n---\n/` be a better approach?

johno:
Yeah, great catch @busypeoples, I've updated the regex.

PR #22, opened 2018-07-31
Fixed issues with carousel dots

felizalde:
Hello, great work on this!
I was playing with this a bit, and I fixed some little issues. 

My changes:
- Removed a unnecessary case clause.
- Fixed the box-shadow property for the dot focused. It only was appearing clicking it and not scrolling with arrows.
    _This was happening:_
img1 img2 
    _now:_
fix
- Update tests to changes.

Do you have some plan/thought about provide different modes to slide (like a vertical-slide, in addition to the current horizontal-slide)? 

Thanks!!

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -165,10 +165,9 @@ export class SlideDeck extends React.Component {
     if (e.metaKey || e.ctrlKey || e.shiftKey || e.altKey) return
     switch (e.key) {
       case 'ArrowRight':
-      case ' ':
         e.preventDefault()
         this.update(inc)
-        break
+        break     
       case 'ArrowLeft':
         e.preventDefault()
         this.update(dec)
--- a/src/index.js
+++ b/src/index.js
@@ -70,7 +70,7 @@ const Dot = styled.button([], {
   },
   props => ({
     opacity: props.active ? 0.5 : 0.125,
-    boxShadow: props.focus ? '0 0 0 1px' : '0 0 0 0px'
+    boxShadow: props.focus ? '0 0 0 1px' : 'none'
   }),
   space,
   color
--- a/src/index.js
+++ b/src/index.js
@@ -66,13 +66,11 @@ const Dot = styled.button([], {
   width: '8px',
   height: '8px',
   color: 'inherit',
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -163,12 +162,12 @@ export class SlideDeck extends React.Component {
 
   handleKeyDown = e => {
     if (e.metaKey || e.ctrlKey || e.shiftKey || e.altKey) return
+
     switch (e.key) {
       case 'ArrowRight':
-      case ' ':
```
jxnblk: Thanks for the PR! This is intentional – it advances with the space bar

src/index.js
```diff
@@ -163,12 +162,12 @@ export class SlideDeck extends React.Component {
...
         e.preventDefault()
         this.update(inc)
-        break
+        break
```
jxnblk: Looks like your editor is adding whitespace?

src/index.js
```diff
@@ -66,13 +66,11 @@ const Dot = styled.button([], {
...
-    outline: 'none',
-    boxShadow: '0 0 0 1px'
-  }
-},
+  outline: 'none'
+  },
   props => ({
-    opacity: props.active ? 0.5 : 0.125
+    opacity: props.active ? 0.5 : 0.125,
+    boxShadow: props.focus ? '0 0 0 1px' : 'none'
```
jxnblk: Would this be confusing to change the browser's focus behavior? Not sure that it would be expected to navigate to a link with the tab key if that's what this is meant to do?

src/index.js
felizalde: Oh cool. I thought that was for something, but not think in the space bar.

src/index.js
felizalde: I think I did not understand you. This only determine the correct value for boxShadow for the button (Dot) with the index of the current slide displayed. The browser's focus behavior was not working well moving slides with the arrow keys.

jxnblk: The tab key and shift+tab should cycle through focusable elements, then return key should select the slide. I believe it's working as expected on my end?

PR #37, opened 2018-08-01
Add Fragments Component

busypeoples:
This is only a proof of concept. It would enable to skip through fragments. 
Currently it would listen to key up and key down (to not clash with key left/right inside the main component).

### DIFF
--- a/docs/index.mdx
+++ b/docs/index.mdx
@@ -28,10 +28,10 @@ import Box from 'superbox'
   # 4 reasons to use Reason
 
   <FragmentBlock>
-    - Based on OCaml
-    - Types
-    - Community
-    - Interop
+    <div>Based on OCaml</div>
+    <div>Types</div>
+    <span>Community</span>
+    <div className="some-effects">Interop</div>
   </FragmentBlock>
 ---
 
--- a/docs/index.mdx
+++ b/docs/index.mdx
@@ -1,5 +1,5 @@
 export { future as theme } from '../src/themes'
-import { Image } from '../src'
+import { Image, FragmentBlock } from '../src'
 import Layout from './Layout'
 
 # mdx-deck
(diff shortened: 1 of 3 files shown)


busypeoples:
Improved how fragments are defined. We don't care how they are blocked.

```
<FragmentBlock>
    <div>Based on OCaml</div>
    <div>Types</div>
    <span>Community</span>
    <div class="some-effects">Interop</div>
</FragmentBlock>
```

jxnblk:
This one's looking great! Is there anything else you'd like to get in for the initial implementation here? I think that hooking into routing and the main app state could be handled as an improvement on top of this later.

I'm a little unsure about the component name... Spectacle calls theirs [Appear](https://github.com/FormidableLabs/spectacle#appear), you have Fragment (which technically makes sense, but seems a little less familiar), and I originally called it Reveal or Step... Have you thought any more about what name makes most sense? The API can always be updated later if we can't figure out a super duper great name right now

busypeoples:
Cool. We can name it whatever would fit best, `steps` sounds very good, but whatever suits best. So we should skip the url updating part? Sure, we can update the handling later on. We wouldn't have to change how this is handled in the markdown, so like you said, we can add it as an improvement.

busypeoples:
@jxnblk One thing we should definitely change is the direction of how we handle the key events. Currently arrow up, jumps to the next step, and arrow down is one step back. I think it should be the other way round. Arrow down is next step. What do you think?

jxnblk:
Oh! Didn't even notice that the arrows were reversed... I think arrow down for next makes most sense, since most of the time that's how it'll be visibly revealed

busypeoples:
I changed that already, forget it. I thought I wanted to change, but just noticed that already works with arrow down for next step.

jxnblk:
For the name, I think I'm leaning more towards `Appear`, since that's what Spectacle calls it and seems sort of inline with some of the language in the Keynote app – you cool with that?

busypeoples:
Sure, whatever is the most obvious!

jxnblk:
I don't mind making the change after merging, but up to you

busypeoples:
Let's merge it, we can still rename it later on.

jxnblk:
Yeah, I'll merge it, I think for initial introduction I'd like to call it Appear, but I'll make that change, and probably add a section to the readme with some docs for the components that can be used in mdx

PR #45, opened 2018-08-03
Add PropType definitions

busypeoples:
Add PropType definitions for documentation purposes.

### DIFF
--- a/src/Dots.js
+++ b/src/Dots.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import { space, color } from 'styled-system'
 import Flex from './Flex'
@@ -22,6 +23,11 @@ const Dot = styled.button([], {
   space,
   color
 )
+Dot.propTypes = {
+  m: PropTypes.number,
+  p: PropTypes.number,
+  bg: PropTypes.string
+}
 Dot.defaultProps = {
   m: 0,
   p: 1,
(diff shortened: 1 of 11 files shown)


src/Dots.js
```diff
@@ -22,6 +23,11 @@ const Dot = styled.button([], {
   space,
   color
 )
+Dot.propTypes = {
+  m: PropTypes.number,
```
jxnblk: All `styled-system` functions have prop types, so this could be:

```
Dot.propTypes = {
  ...space.propTypes,
  ...color.propTypes
}
```

src/Flex.js
```diff
@@ -13,4 +14,8 @@ const Flex = styled.div([], {
   color
 )
 
+Flex.propTypes = {
+  css: PropTypes.object
```
jxnblk: This can also be a string

src/Image.js
```diff
@@ -14,6 +15,11 @@ const Image = styled.div([], {
   props => props.css
 )
 
+Image.propTypes = {
+  width: PropTypes.string,
```
jxnblk: Since styled-system accepts numbers, arrays, or strings, it's probably better to use the library's prop types like above

src/Dots.js
busypeoples: Should we only use the space and color propTypes? Or should we also declare `m` etc.

jxnblk: Those both include all prop types, the space propTypes are here: https://github.com/jxnblk/styled-system/blob/master/src/space.js#L102-L117

jxnblk:
This looks great, thanks!

PR #46, opened 2018-08-03
Add countdown mode for  presentation timer

busypeoples:


busypeoples:
See https://github.com/jxnblk/mdx-deck/issues/47

This will need some styling. I think it's better if you take care of the look and feel @jxnblk.

We can also change the color of the timer when nearing to 0. Currently it only changes the colors when in overtime.

jxnblk:
Awesome! Let me pull this down and check out what it looks like/think a little about the API... might also be nice to support setting a default time in the mdx file itself – i.e. if I'm doing a talk and making a presentation, the time is going to generally be set

busypeoples:
Sure, this is an initial idea. We definitely improve this, especially the look and feel. Excellent idea setting the time via mdx. This would mean we can access the defined time via a context provider instead the randomly defined default time of 30 minutes.

src/Timer.js
```diff
@@ -58,16 +90,43 @@ class Timer extends React.Component {
...
           </Button>
         )}
+
+        {directionMode === direction.countdown && (
+          !on
+            ? <input
+                value={countdownTime/60}
+                onChange={this.updateCountdownTime}
+              />
+            : <Mono px={2}>Presentation Time: {countdownTime/60} minutes</Mono>
```
jxnblk: Might be nice if the input and the running countdown are the same piece of UI – i.e. when it's not running, it's editable as an input; and when it's running it's readonly – would have to make a decision about what happens on pause in that case, but just spitballing here

src/Timer.js
```diff
@@ -58,16 +90,43 @@ class Timer extends React.Component {
...
         {!on && seconds > 0 && (
           <Button mr={1} onClick={this.reset}>
             reset
           </Button>
         )}
+
+        {directionMode === direction.countdown && (
+          !on
+            ? <input
+                value={countdownTime/60}
```
jxnblk: Could be a `type='number'` input if it's a number value

src/Timer.js
```diff
@@ -23,19 +23,51 @@ Button.defaultProps = {
...
+const DEFAULT_COUNTDOWN = 1800
+
 class Timer extends React.Component {
   state = {
     on: false,
     time: new Date().toLocaleTimeString(),
-    seconds: 0
+    seconds: 0,
+    directionMode: direction.normal,
+    countdownTime: DEFAULT_COUNTDOWN
```
jxnblk: If this was initially set from props, would allow the mdx file to pass a value in from the main app

busypeoples: Exactly, so we if have access to this value from mdx, we can replace this easily.

src/Timer.js
busypeoples: Yes, it was type number at some point and makes sense.

src/Timer.js
busypeoples: Yes, this would be actually excellent, making it read-only when running would make a lot of sense.

busypeoples: We could also use an actual toggle button for switching between the modes.

src/Timer.js
busypeoples: So f.e. inside the mdx we would export a `time` or `presentationTime` value? And we can access it via `withDeck` for example? or would we define some sort of `settings` object similar to `theme`

jxnblk: I think e.g. `export const timer = 30` would be nice to have in the mdx file, doesn't need to be a full settings object yet

jxnblk:
Still wanna get this in btw, just need some time to figure out the UI

trevordmiller:
Looking forward to this one. Will be nice to not have to fiddle with a separate timer app or forget the time and go over/under on accident when presenting.

jxnblk:
Thanks for all the help with your PRs and sorry I never got this merged! I'm going to close this and #88 out for the time being as I work on v2

Issue #29, opened 2018-07-31
Syntax highlighting


jxnblk:
- [x] React component
- [x] Markdown code fences 
- [x] Should be themeable
- [x] Include JSX language by default
- [x] Add docs

tomByrer:
[conorhastings/react-syntax-highlighter](https://github.com/conorhastings/react-syntax-highlighter) : Uses Lowlight, powerful, good docs, but requires a bit of configuration to have small file footprint

[akiran/react-highlight](https://github.com/akiran/react-highlight) : uses highlight.js, & also requires extra work to not include every language & skin.  Seems more lightweight, but docs are sparse

[Vamche/react-syntax-highlight](https://github.com/vamche/react-syntax-highlight) : also uses highlight.js, decent documentation, but unknown to React community so not many issues/PR/versions.

jxnblk:
That first one looks great, thanks!

To clarify the theming bullet point, I'd imagine that the syntax theme could be added to an mdx-deck theme:

```js
// example theme
import { docco } from 'react-syntax-highlighter/styles/hljs'

export default {
  highlight: docco
}
```

Then automatically added to the syntax highlight component (and fenced code blocks) using the `withTheme` HOC

busypeoples:
Is there anything missing here or can we close this issue?

PR #53, opened 2018-08-04
Add code highlighting

busypeoples:
https://github.com/jxnblk/mdx-deck/issues/29


busypeoples:
Initial implementation. Uses prism and the supported languages and styles. There is a default style which can be overridden inside the mdx file. Currently two languages are available by default (JavaScript and Elixir) and we can include all supported languages when needed.

src/Code.js
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  getLangauge = (lang) => {
+    return lang ? lang.replace('language-', '') : 'javascript'
+  }
+
+  render() {
+    const { className, children, theme } = this.props
+    const language = this.getLangauge(className)
+    const style = (theme && theme.highlight) ? theme.highlight : defaultTheme
```
jxnblk: I think `theme` will always be an object, even if it's an empty object

src/components.js
```diff
@@ -155,7 +156,7 @@ const code = props => {
         </Notes>
       )
     default:
-      return <Pre {...props} />
+      return <Code {...props} />
```
jxnblk: Would be good to keep the basic pre styles/theming intact for the new component – maybe it renders a default `<pre>` tag with no syntax highlighting if there's no `highlight` field in the theme?

src/Code.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { withTheme } from 'styled-components'
+import SyntaxHighlighter, { registerLanguage } from "react-syntax-highlighter/prism-light"
+import defaultTheme from 'react-syntax-highlighter/styles/prism/atom-dark'
+import javascript from 'react-syntax-highlighter/languages/prism/javascript'
+import elixir from 'react-syntax-highlighter/languages/prism/elixir'
```
jxnblk: JS I think makes sense as a default (since this is all built with React/JS), but elixir seems a little opinionated. I'd like the languages to be configurable somehow, but not sure the best way to approach that with the way `react-syntax-highlighter` works...

src/Code.js
busypeoples: I think one of the tests failed before hand, that's why I changed this to a more defensive approach. We will also need to update the tests.

docs/index.mdx
```diff
@@ -1,6 +1,7 @@
 export { future as theme } from '../src/themes'
-import { Image, Notes, Appear } from '../src'
+import { Image, Notes, Appear, Code } from '../src'
 import Layout from './Layout'
+export highlight from 'react-syntax-highlighter/styles/prism/solarizedlight'
```
jxnblk: This should be in the theme, shouldn't it?

src/components.js
busypeoples: Makes sense.

src/Code.js
busypeoples: I just randomly added a second language, just wanted to choose a second language.

docs/index.mdx
busypeoples: This is an overriding example.

src/Code.js
jxnblk: Yeah, some of the theming context stuff will fail if the provider or a default prop isn't there, not sure if that's related or not

src/entry.js
```diff
@@ -4,14 +4,14 @@ import SlideDeck from './index'
...
-const { theme, components, Provider } = mod
+const { theme, highlight, components, Provider } = mod
 
 export default class App extends React.Component {
   render () {
     return (
       <SlideDeck
         slides={slides}
-        theme={theme}
+        theme={{...theme, highlight}}
```
jxnblk: Ah, I see what's happening here now... Do you think it's even worth supporting this outside the theme? I'm starting to think that options like `components` and `Provider` might also make more sense to be in the same `theme` object 🤔

src/Code.js
busypeoples: I think the theme context is undefined initially, but will recheck.

src/entry.js
busypeoples: Maybe we can group them. What would the advantage be?

src/components.js
busypeoples: Just checked, we only know when inside the `Code` component. The problem is that if there is no `highlight` field we fall back to the default theme which is defined at 
https://github.com/busypeoples/mdx-deck/blob/fe63e5b7d67c47fdc5522409b51e279a3636173a/src/Code.js#L5

I choose a random default theme, but we could use the prism default f.e.

jxnblk: Yeah, if this `code` component was wrapped with `withTheme` then it should have access to the theme, right?

busypeoples: On the other hand this means themes have to be defined explicitly inside an mdx, would we even need a default theme inside `Code` then?

src/entry.js
jxnblk: This is totally a non-blocker, just thinking, but the idea being to simplify the API a bit and make themes a little more portable by including most of the reusable configuration anyone might want to add

jxnblk:
Sorry for all the questions/discussion here, I can totally take what you've started and continue with it if you don't have time, but this is generally what I'm thinking the docs for this could look like – hopefully this helps clarify a little bit of what I'm thinking from an end user standpoint:



## Syntax Highlighting

By default fenced code blocks do not include any syntax highlighting.
Syntax highlighting in fenced code blocks can be enabled by providing either a `highlight` or `prism` style object in a theme.

```js
// example theme
import { theme } from 'mdx-deck/themes'
import prismStyle from 'react-syntax-highlighter/styles/prism/solarizedlight'

export default {
  ...theme,
  prism: {
    style: prismStyle
  }
}
```

Now any fenced code block in your mdx file will be highlighted:

````mdx
export { default as theme } from './theme'

This is highlighted:

```js
const foo = 'hello'
```

````

By default, only JavaScript is enabled for syntax highlighting, to keep bundle sizes to a minimum.
To enable another languages, add a `languages` object to the `prism` or `highlight` object in the theme.

```js
// example theme
import { theme } from 'mdx-deck/themes'
import prismStyle from 'react-syntax-highlighter/styles/prism/solarizedlight'
import prismHTML from 'react-syntax-highlighter/languages/prism/html'

export default {
  ...theme,
  prism: {
    style: prismStyle,
    languages: {
      html: prismHTML
    }
  }
}
```

jxnblk:
With the approach above, I think the default `code` component can be wrapped with the `withTheme` provider to check whether syntax highlighting is enabled and either render the basic one or the Code component with highlighting. There should always be a default theme defined (https://github.com/jxnblk/mdx-deck/blob/master/src/index.js#L51), but there might be some bugs/edge cases I've missed.

I *think* the Code component that you've added could call `registerLanguage` in the constructor based on the API, but I might also be missing something there...

busypeoples:
Changes the behaviour. Maybe take another look. +1 for the great idea of enabling to define further languages via the theme.

docs/theme.js
```diff
@@ -0,0 +1,13 @@
+import { future as theme } from '../dist/themes'
+import prismStyle from 'react-syntax-highlighter/styles/prism/solarizedlight'
+import prismJsx from 'react-syntax-highlighter/languages/prism/jsx'
```
jxnblk: Hmm, wonder if jsx should be enabled by default as well, seems like there's a high likelihood of using this for react/jsx

busypeoples: Yes, makes sense actually!

test/components.js
```diff
@@ -5,7 +5,8 @@ import components from '../src/components'
 
 const blacklist = {
   pre: true,
-  img: true
+  img: true,
+  code: true
```
jxnblk: Was this causing errors or something?

busypeoples: I think so. I'll verify.

busypeoples: It seem we need to initialize the theme provider,

test/components.js
```diff
@@ -13,7 +14,7 @@ test.each(
     .filter(key => !blacklist[key])
     .map(key => {
       const Component = components[key]
-      return [ key, <Component /> ]
+      return [ key, <ThemeProvider theme={{}}><Component /></ThemeProvider> ]
```
jxnblk: Yeah, the other thing to do for components that rely on the theme is to provide it as a default prop, but this works as well

jxnblk:
This is looking great, should we :shipit: ?

Issue #51, opened 2018-08-03
Hook Appear component into state/routing


jxnblk:
Related to #34

tamouse:
Not sure if this is related, but `<Appear>` doesn't seem to work in from presentation mode:

- on the speaker (presentation) screen, it works as expected
- on the audience screen, the steps aren't revealed.

p.s. I love this app so much! Way more fun than `reveal.js`

PR #54, opened 2018-08-04
Fix Appear handling

busypeoples:
Check if slide active status has changed and reset the steps if the status has changed. Extend `withSlide` to also provide the slide active status.

### DIFF
--- a/src/Appear.js
+++ b/src/Appear.js
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { withSlide } from './Slide'
+import { incStep, decStep } from './index'
 
 export default withSlide(class Appear extends React.Component {
   static propTypes = {
@@ -24,13 +25,11 @@ export default withSlide(class Appear extends React.Component {
     switch (e.key) {
       case 'ArrowDown':
         e.preventDefault()
-        update(state => ({
-          step: state.step < children.length - 1 ? state.step + 1 : state.step
-        }))
+        update(incStep(children))
         break
       case 'ArrowUp':
         e.preventDefault()
-        update(state => ({ step: state.step >= 0 ? state.step - 1 : -1 }))
+        update(decStep())
         break
     }
   }
--- a/src/Appear.js
+++ b/src/Appear.js
@@ -1,9 +1,11 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import { withSlide } from './Slide'
 
-export default class Appear extends React.Component {
+export default withSlide(class Appear extends React.Component {
   static propTypes = {
-    children: PropTypes.array.isRequired
+    children: PropTypes.array.isRequired,
+    slide: PropTypes.object.isRequired
   }
 
   constructor(props) {
@@ -16,6 +18,12 @@ export default class Appear extends React.Component {
 
   componentDidMount() {
     document.body.addEventListener('keydown', this.handleKeyDown)
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.props.slide.active !== prevProps.slide.active) {
+      this.setState(state => ({ fragmentStep: -1}))
+    }
   }
 
   componentWillUnmount() {
@@ -71,4 +79,4 @@ export default class Appear extends React.Component {
       </React.Fragment>
     )
   }
-}
+})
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -184,14 +198,18 @@ export class SlideDeck extends React.Component {
...
                 <GoogleFonts />
                 <Carousel index={index}>
                   {slides.map((Component, i) => (
                     <Slide
                       key={i}
                       id={'slide-' + i}
-                      index={i}>
+                      index={i}
+                      active={i === index}
+                      update={this.update}
```
jxnblk: I think this is a little confusing as to what's happening here? This would provide the main app's state update methods in two different contexts now, right?

src/Appear.js
```diff
@@ -24,51 +18,46 @@ export default class Appear extends React.Component {
...
         e.preventDefault()
-        this.setState(state => {
-          return {
-            fragmentStep:
-              state.fragmentStep < state.fragments.length - 1
-                ? state.fragmentStep + 1
-                : state.fragmentStep
-          }
-        })
+        update(state => ({
```
jxnblk: I think there's two different approaches to this state updating pattern that I don't mind, either:
1. Passing a generic `update` method down and importing "updater" functions from a common module, e.g. `update(decrementStep)` (not sure about the verb)
2. Passing more explicitly names methods down, e.g. `this.props.deck.decrementStep`

busypeoples: Sure, what would be the preferred way? Maybe passing down `update` might have an advantage of enabling custom components to be written. For now, we can also go for the second approach and think about a more generic approach when we really needed it.
Either way it's definitely a good idea to move this functions into a common module.

busypeoples:
These changes will also solve https://github.com/jxnblk/mdx-deck/issues/51

src/index.js
busypeoples: What other context would we have available? Might have missed something here.

jxnblk: Yeah, sorry this is moving around quickly, the main app uses this context module https://github.com/jxnblk/mdx-deck/blob/master/src/context.js here https://github.com/jxnblk/mdx-deck/blob/master/src/index.js#L175

jxnblk: Can `import { withDeck } from './context'` to hook into it

jxnblk: The slide context is because the notes need to know which slide index to appear in – there might be a way to consolidate this all into the slide context tbh

src/Appear.js
jxnblk: Yeah, I'm okay with trying 1 to start with – it gives you a little more flexibility

src/index.js
```diff
@@ -111,8 +126,13 @@ export class SlideDeck extends React.Component {
   }
 
   handleStorageChange = e => {
-    const index = parseInt(e.newValue, 10)
-    this.setState({ index })
+    if (e.key === MDX_SLIDE_INDEX) {
+      const index = parseInt(e.newValue, 10)
+      this.setState({ index })
+    } else if (e.key === MDX_SLIDE_STEP, 10) {
```
jxnblk: 👍

src/Appear.js
```diff
@@ -1,17 +1,14 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import { withDeck } from './context'
+import { withSlide } from './Slide'
+import { incStep, decStep } from './index'
```
jxnblk: Cool, I can pull all these out into a dedicated `updaters` module after merging

test/index.js
```diff
@@ -36,9 +36,9 @@ describe('components', () => {
   transition-timing-function: ease-out;
   -webkit-transition-duration: .3s;
   transition-duration: .3s;
-  -webkit-transform: translateX(NaN%);
-  -ms-transform: translateX(NaN%);
-  transform: translateX(NaN%);
```
jxnblk: whoops!

jxnblk:
LGTM - is there anything else you wanted to do here or is it good to go?

test/index.js
busypeoples: ìndex` was missing inside the Carousel test.

busypeoples:
We might need to be more defensive with how we update the steps part in the url. But I can improve that via an own issue.

PR #56, opened 2018-08-04
Adjust build and add contributing doc

jxnblk:


jxnblk:
Heads up @busypeoples this is going to slightly change local development, but I think it'll be better for the package setup

PR #64, opened 2018-08-04
Make jsx default and add a ruby example

busypeoples:


src/Code.js
```diff
@@ -4,6 +4,7 @@ import { withTheme } from 'styled-components'
 import SyntaxHighlighter, { registerLanguage } from "react-syntax-highlighter/prism-light"
 import defaultTheme from 'react-syntax-highlighter/styles/prism/atom-dark'
 import javascript from 'react-syntax-highlighter/languages/prism/javascript'
+import jsx from 'react-syntax-highlighter/languages/prism/jsx'
```
jxnblk: Whoops! I did this in my branch as well...

docs/index.mdx
```diff
@@ -39,6 +39,10 @@ import Box from 'superbox'
 <button>code example</button>
 ```
 
+```ruby
+puts "This is a Ruby Example"
+```
```
jxnblk: I think I'll probably do a complete pass through this demo doc and change things quite a bit, so feel free to use it for development, but don't worry about having examples of all the features in here for the time being

src/Code.js
busypeoples: All good. I'll close this PR.

Issue #63, opened 2018-08-04
Add support for custom `components` and `Provider` to themes


jxnblk:
Instead of separate exports, custom `components` and `Provider` could be added to a theme – this makes themes a little more powerful and hopefully simplifies the API a bit.

```js
// example
import { theme } from 'mdx-deck/themes'
import Provider from './Provider'
import Heading from './Heading'

const theme = {
  ...theme,
  font: 'Georgia, serif',
  Provider,
  components: {
    h1: Heading
  }
}
```

PR #71, opened 2018-08-04
Move external provider and components into theme

busypeoples:


src/index.js
```diff
@@ -200,13 +198,16 @@ export class SlideDeck extends React.Component {
...
-      components,
-      Provider,
       width,
       height
     } = this.props
     const { index, length, mode, step} = this.state
 
+    const {
+      components = {},
+      Provider = DefaultProvider
```
jxnblk: Nice!

src/entry.js
```diff
@@ -4,16 +4,14 @@ import SlideDeck from './index'
 
 const mod = require(DOC_FILENAME)
 const slides = mod.default
-const { theme, components, Provider } = mod
```
jxnblk: This looks like it'd mean a breaking change – what do you think of keeping backwards compatibility even if it's not documented?

busypeoples: See what you mean. Then let's keep it.

busypeoples:
No idea why this doesn't pass. Double checked that `DefaultProvider` is not declared twice. This is interesting.

jxnblk:
I suspect it's because of this with the merge: https://github.com/jxnblk/mdx-deck/blob/master/src/index.js#L8

jxnblk:
Awesome, LGTM!

Fixes #63

Issue #68, opened 2018-08-04
Make Appear work with left/right keys


busypeoples:
To simplify things, stepping through the steps inside an `Appear` component should happen via left/right arrows, just like stepping through regular slides.

PR #75, opened 2018-08-05
Enable components to subscribe to events

busypeoples:
See #68 

An initial approach to improve events handling


busypeoples:
We can improve this, but this initial approach would enable us to subscribe to events from within components.

jxnblk:
Interesting, I'm generally not a huge fan of pub/sub in React, so might need to think on this a bit. What does this give us that can't be achieved with React state and context?

busypeoples:
We want to keep the event handling centralized I would think. Currently this is handled inside `SlideDeck`, which makes a lot of sense. But how can we hook into the event from outside `SlideDeck`. Maybe I'm missing something here again, how would you enable to also listen to left/right keys f.e.? Our `Appear` component for example doesn't really know anything about what left / right keys do outside it's context, but it needs to take over control for revealing the steps.

jxnblk:
Right, I don't love what I have in there currently, but the Notes component sort of registers itself with the parent state by setting the `notes` object. Technically you should be able to loop over `props.children` to know what's being rendered where, but the `addNotes` method was just a quick way of handling that

busypeoples:
What would this imply? We would need some form of introspection if I understand correctly?

busypeoples:
As a first step we could make this work like the `Notes` component, just to make this work with left/right keys. What do you think?

busypeoples:
Changed the handling, this works like the `Notes` component, which is definitely better than the previous approach with the event subscription. This also makes the `Appear` more simple, as it doesn't have to take care of any key events.

tomByrer:
Could this be used to auto-advance slides, like a carousel?  Even advancing via remote API would be nice, though keyboard would be a workaround.

busypeoples:
@tomByrer Yes, it would enable any auto advancing and other forms of advancing. The auto-advancing feature sounds very useful. Maybe create specific issues and let's see what @jxnblk thinks about these features.

jxnblk:
@busypeoples thanks for all your help with these PRs! I'm closing this one in favor of #144, but going to try to get the other two open PRs merged in soon

PR #84, opened 2018-08-06
Add note about `.babelrc` file

rafaelrinaldi:
When I first tried out the project I was getting a weird error because I didn't have a local `.babelrc` file on the project. Would be helpful if that was listed as a setup requirement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,12 @@ pointing to the `.mdx` file to start the dev server:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


jxnblk:
It *shouldn't* require a `.babelrc`, that's definitely a bug. Would you mind opening an issue with some details about your setup? i.e. node and npm versions, other installed dependencies, etc?

rafaelrinaldi:
Sure thing.

PR #85, opened 2018-08-07
Use template literals

neighborhood999:


src/Zoom.js
```diff
@@ -6,8 +6,8 @@ const ZoomRoot = styled.div([], {
   backgroundColor: 'white',
 },
   props => ({
-    width: (100 * props.zoom) + 'vw',
-    height: (100 * props.zoom) + 'vh',
+    width:  `${(100 * props.zoom)}vw`,
```
jxnblk: Thanks, but I'm not sure this gains much from a readability standpoint...

Issue #87, opened 2018-08-08
Enable Auto-Transition


busypeoples:
Add a settings option to define auto-transitions.
The `autoTransition` constant needs to define the transition time in seconds, i.e.

```js
// index.mdx

export const transitionTime = 5

```
Also see comment by @tomByrer in https://github.com/jxnblk/mdx-deck/pull/75

busypeoples:
A possible documentation:

## Auto Transition

To enable the automatic transitioning of your slides and appears, export a `autoTransition` constant.
`autoTransition` should set the time between transitions.
For example if you need to transition every 10 seconds, add the following to your `MDX` file:

```js
// Example index.mdx

export const autoTransition = 10
```

PR #88, opened 2018-08-08
Auto transition

busypeoples:
Based on #75 

Also see #87


README.md
```diff
@@ -243,15 +243,25 @@ This shows a list of all slides on the left and a preview of the current slide o
...
+## Auto Transition
+
+To enable the automatic transitioning of your slides and appears, export a `autoTransition` constant.
+`autoTransition` should set the time between transitions.
+For example if you need to transition every 10 seconds, add the following to your MDX file
+
+```js
+// Example index.mdx
+
+export const autoTransition = 10
```
jxnblk: Seems like a legitimate feature to add, but I'm not sure about it being configured in the mdx document. My thinking is that the configuration in mdx should be things that you *always* want set, whereas this feature seems more like something to enable in the app state – maybe via a button in presenter mode or a keyboard shortcut...

src/index.js
```diff
@@ -85,15 +86,18 @@ export class SlideDeck extends React.Component {
...
+    autoTransition: -1
   }
 
   state = {
     length: this.props.slides.length,
     index: 0,
     mode: modes.normal,
     notes: {},
-    step: -1
+    fragments: {},
```
jxnblk: NB: it seems like it might be good to combine notes/fragments/etc into a single object for the current slide, but can definitely be a followup thing

src/index.js
```diff
@@ -105,31 +109,32 @@ export class SlideDeck extends React.Component {
...
     if (e.metaKey || e.ctrlKey || e.shiftKey) return
+
     const alt = e.altKey
 
     switch (e.keyCode) {
       case keys.right:
       case keys.space:
         e.preventDefault()
-        this.update(inc)
+        this.next(this.state)
```
jxnblk: Any reason to switch away from the functional set state approach here?

busypeoples: Wanted to decouple the event handling from the actual state updating. Moved `this.update(...)` into `next`.
But yes, no need to pass the current state, we can access it inside `next` directly.

README.md
busypeoples: Sure, will move this outside of the configuration.

tomByrer: Yes, I like the idea of keeping the `slides.mdx` just for content & (basic-ish) layout, & deep configs external.

busypeoples:
I'm keeping this branch in sync. No idea if it's still interesting. Moved the auto transition setting into the presenter mode. Would still need some proper styling probably.

jxnblk:
Yeah, still wanna get this in, just been traveling the past week – thanks for keeping this up to date!

PR #90, opened 2018-08-08
Fixes missing Dots and navigation buttons on initial load when in normal mode.

drmikeh:
The mode was being set to undefined in the getMode function when initially loading the application with a normal mode. getMode reads the mode URL parameter looking for a value but the parameter is not there when in normal mode.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -150,7 +150,7 @@ export class SlideDeck extends React.Component {
   }
 
   getMode = () => {
-    const { mode } = querystring.parse(window.location.search.replace(/^\?/, ''))
+    const { mode } = querystring.parse(window.location.search.replace(/^\?/, '')) || 'normal'
     this.setState({
       mode: modes[mode]
     })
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -150,7 +150,7 @@ export class SlideDeck extends React.Component {
   }
 
   getMode = () => {
-    const { mode } = querystring.parse(window.location.search.replace(/^\?/, ''))
+    const { mode } = querystring.parse(window.location.search.replace(/^\?/, '')) || 'normal'
```
jxnblk: Thanks! It'd be good to add some tests for this sort of thing. Based on looking at the code, does this work as expected?

busypeoples: Wouldn't it be better to be explicit about it and only update `mode` when an a mode is actually defined. `mode` has a default when the state is defined.

jxnblk:
I think #94 should mostly fix this issue, unless someone explicitly sets the query string to `?mode=`

jxnblk:
Thanks for this, but going to go with #101 instead

PR #91, opened 2018-08-08
[Feature] Mobile Swiping support

EmaSuriano:
Hi to everybody! 👋 

I've been using mdx-deck for a few days, I really love how fast is to create slides now! One thing that I noticed the first time I saw it (I was using my phone), was that the slides are practically unusable for mobile users (they don't have any way to move forward). 

For that reason, I created this PR! Basically, I implemented [react-swippeable](https://github.com/dogfessional/react-swipeable) which is a really good component for detecting swipes (works in Mobile and Web 😄 ).

This is a recording of how it's working, simulating a mobile with the chrome dev tools.

ezgif-3-0539f035c2

Regarding testing, I updated the snapshot of the SlideDeck rendered. Please tell me if you want to perform another kind of testing? 

Feel free to request any change! I hope this feature will improve this library 👍


jxnblk:
Awesome thanks! I might need a few days to check this out, but generally looks great. I would like to have more functional tests in for this sort of thing, but that can always be added later. Would you mind removing the `package-lock.json` in the meantime?

EmaSuriano:
Yeah sure! But don't you want to leave the new depency inside of it? I mean it's the purpose of the `package-lock.json`

jxnblk:
See https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514

It’s more important to have a lock file in the end user’s package, to ensure they have the exact versions of the libraries they’d like to use along with mdx-deck

EmaSuriano:
Wow, I didn't know about that, really good explanation :clap: I just removed it!  What about adding the `package-lock.json` to the  `.gitignore` file?

jxnblk:
This looks great! Thanks for putting this together!

yduman:
Is this feature on master already? I did a rewrite of my homepage yesterday using mdx-deck. But the swipe behavior is inverted somehow. When I swipe right, it shows me the next slide, but this should happening when swiping left.

PR #105, opened 2018-08-14
Correct named import of theme to default import for custom themes

kyleshevlin:
When I was attempting to create a custom theme and following the example, I noticed that only the new styles I was seeing were the ones I had added. No extension of the base theme was taking place. I logged out the `theme` object I was importing and found it was `undefined`. This is because `theme` is not a named export (at least not under that name). The `base` theme is exported under the `default` name in the distributed code. Thus, I corrected the documentation here to reflect the theme as a default export. This should also be done for the theme documentation, which I will do as a follow up to this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ It's recommended that all custom themes extend the default theme as a base.
 
 ```js
 // example theme.js
-import { theme } from 'mdx-deck/themes'
+import theme from 'mdx-deck/themes'
 
 export default {
   // extends the default theme


kyleshevlin:
Just want to point out while this looks similar to #106, this is for the README, that one was for the theming documentation. I didn't recognize at the time the need to update both concurrently. Forgive me, it was late at night 😀

jxnblk:
Thanks! Sorry, was trying to merge from my phone and had to rerun the CI tests and forgot to hit merge

PR #106, opened 2018-08-14
Corrects named export to default export in theming documentation

kyleshevlin:
Just as in PR #105, I have updated the docs to correctly import the base theme as a default import.


jxnblk:
Awesome, thanks!

PR #117, opened 2018-08-18
Add link to repository

alexkrolick:
Adds a field to package.json so a link appears to the source on NPM

image


jxnblk:
Thanks!

Issue #112, opened 2018-08-15
custom webpack configurations.


kwangkim:
Currently, `cli.js` contains setup for webpack.
If someone want to add extra remark mdPlugins or webpack loaders, user may need to modify `cli.js` manually. 

Do you consider to consider options to allow users add extra mdPlugins or webpack loaders?

mmmeff:
Would love this as well, this is necessary for adding typescript support.

For an implementation example, see how https://github.com/pedronauck/docz implemented support for a `typescript: true` flag in their configuration as well as custom webpack loader options.

jxnblk:
Yeah, I'll likely add support for custom `webpack.config.js` from the CLI and would accept a PR for this

connor-baer:
Hey @jxnblk, I'd love to give this a shot. I can see two options for implementing support for custom webpack configs: 

- **Automatically use a `webpack.config.js` file if it exists.** Pros: easy to use, very flexible. Cons: Overrides default webpack config fully, so the default config has to be reimplemented & kept up to date.
- **Replicate Next.js' approach:** A custom config file that exports a function which receives the default config as an argument and returns the modified config. Pros: No need to reimplement the default config. Cons: less intuitive.

WDYT? Which approach do you prefer or did you have something completely different in mind?

jxnblk:
I think a good approach is to:

1. Add a `--webpack` option to the CLI to point to a custom config
2. Automatically look for a `webpack.config.js` from the CLI using `find-up` as [done here in x0](https://github.com/c8r/x0/blob/master/cli.js#L114-L119)
3. Use [webpack-merge](https://www.npmjs.com/package/webpack-merge) to merge with the built-in config, making sure `mdx-deck/loader` is still used

PR #136, opened 2018-08-25
Add support for custom webpack config

connor-baer:
Based on #112. 

## Changes

1. Add a `--webpack` option to the CLI to point to a custom config
2. Automatically look for a `webpack.config.js` from the CLI using `find-up` as [done here in x0](https://github.com/c8r/x0/blob/master/cli.js#L114-L119).
3. Use `webpack-merge` to merge with the built-in config

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -2,6 +2,7 @@ const path = require('path')
 const webpack = require('webpack')
 const HTMLPlugin = require('mini-html-webpack-plugin')
 const ProgressBarPlugin = require('progress-bar-webpack-plugin')
+const merge = require('webpack-merge')
 const chalk = require('chalk')
 const remark = {
   emoji: require('remark-emoji'),
@@ -107,7 +108,7 @@ const baseConfig = {
 }
 
 const createConfig = (opts = {}) => {
-  const config = Object.assign({}, baseConfig)
+  const config = merge(baseConfig, opts.webpack)
   config.context = opts.dirname
 
   config.resolve.modules.push(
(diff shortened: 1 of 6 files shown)


connor-baer:
I'm on the road and my laptop is running out of battery, will test and finish this later.

connor-baer:
@jxnblk This is ready for review. 

I wasn't quite sure how to make sure `mdx-deck/loader` is still used. The original ticket asks for a way to add additional remark mdPlugins, so those would have to be merged in somehow. Maybe a note in the documentation would suffice?

jxnblk:
Thanks! This looks great!

> I wasn't quite sure how to make sure mdx-deck/loader is still used.

Yeah, maybe that's something to note in the docs, but I'm sure people will break the setup. Could be fine just mentioning this in the docs.

lib/config.js
```diff
@@ -107,7 +108,7 @@ const baseConfig = {
 }
 
 const createConfig = (opts = {}) => {
-  const config = Object.assign({}, baseConfig)
+  const config = merge(baseConfig, opts.webpack)
```
jxnblk: I can't remember, but do you know if this clones the `baseConfig`?

connor-baer: Looks like it: https://github.com/survivejs/webpack-merge/blob/master/src/index.js#L15

connor-baer:
I've added a note about the default `mdx-deck/loader`.

jxnblk:
Looks great, thanks! 🎉

PR #142, opened 2018-08-27
Fix slide index in presenter view

connor-baer:
Related to #122.

Slide number in presenter mode is using `index` so it shows `0` to `n-1` but should be `1` to `n`.


jxnblk:
Thanks! 🎉

PR #144, opened 2018-08-28
Context refactor

### LINKED ISSUE #68: Make Appear work with left/right keys

jxnblk:
- Consolidates state into a single context
- Adds support for stepping through Appear steps with left/right arrow keys and space bar - fixes #68 
- Reorganizes modules
- Adds tests for updater functions
- Simplifies Appear component
- Adds `metadata` state for notes, steps, etc.
- Exports `withDeck` HOC for context consumption

cc @pomber: Want to be sure you're aware of some of these internal changes. Hopefully this will make the code-surfer integration simpler, but also might run into some breaking changes. I can help out/clarify the changes over the next few days to make sure nothing breaks


pomber:
I like it. Much simpler.  
CodeSurfer component is similar to Appear so I think I won't have problems adapting it.

jxnblk:
@pomber if you're cool with the proposed changes here, I might go ahead and roll this out, but if it's helpful for me to publish a prerelease for testing first I could do that. Also happy to help with any missing docs or updates that might be helpful for CodeSurfer

pomber:
A prerelease would be nice :). I'll try to test the new integration tonight.

jxnblk:
Cool, it's published as `mdx-deck@alpha` (v1.7.0-0) – I'm guessing the Appear component will serve as a reference, but feel free to ask any questions or let me know how it goes

pomber:
It mostly worked.  
I'm having some minor issues with `deck.step` when going backwards from one slide to the previous one. 

I'm implementing sCU like this:
```js
    shouldComponentUpdate() {
      return this.props.deck.active;
    }
```
but I'm not sure if it's the correct way to do it because I'm missing some updates when the slide changes.

Anyway, if you want to release 1.7 go ahead, I'll publish a new verion of CodeSurfer afterwards.
Thanks!

PR #150, opened 2018-08-30
Add .gitignore to basic template

macklinu:
When running `npm init deck path-to-deck`, the initial commit contains the entire `node_modules/` directory. This PR adds a `.gitignore` file to ignore that and the generated build files (`dist/`).


jxnblk:
Awesome, thanks!!

PR #158, opened 2018-09-11
Use translate3d

quietshu:
Use `translate3d` instead of `translateX` for performance reason.

https://stackoverflow.com/questions/22111256/translate3d-vs-translate-performance


quietshu:
@jxnblk I just ran that jsperf tests using latest Chrome, it seems that there's a difference for sure: https://jsperf.com/translate3d-vs-xy. 😄

jxnblk:
Interesting. Barely makes a difference in Safari. Mind updating the snapshots to get the tests passing?

PR #160, opened 2018-09-13
Specify host cli option

silasb:
I normally do all my JS coding from my home server.  This PR enables to
to run the dev + hot client server correctly.


jxnblk:
Thanks! Seems like a sensible addition

PR #161, opened 2018-09-14
Shrink the dot container width

jesstelford:
This reduces the surface area that becomes unclickable within the window (eg; when a custom layout sticks a footer to the bottom of the viewport).

PS: I'm pretty sure this breaks the tests, but my Jest-fu isn't strong enough to figure out how to rectify it - I'd love some help there!

### DIFF
--- a/src/Provider.js
+++ b/src/Provider.js
@@ -6,9 +6,9 @@ import { modes } from './constants'
 
 const Bottom = styled.div([], {
   position: 'fixed',
-  left: 0,
-  right: 0,
   bottom: 0,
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks! I'll try to take a look over the weekend. I think the tests are failing because it's doing a huge snapshot (which I should replace with more functional tests)

jxnblk:
I'm not entirely sure what this change is meant to do, tbh. Can you clarify what your intent with this is? I know there's an issue when dots overflow, but failing to see what this does to help

PR #162, opened 2018-09-14
Allow absolute positioning within a slide

jesstelford:
Without this, anything that is given `position:absolute; left: 0` gets put at the left edge of the carousel, not the individual slide.

### DIFF
--- a/src/Slide.js
+++ b/src/Slide.js
@@ -13,6 +13,7 @@ const Root = styled.div([], {
   overflow: 'hidden',
   width: '100%',
   height: '100%',
+  position: 'relative',
   '@media print': {
     width: '100vw',
     height: '100vh',


jxnblk:
Thanks! We probably shouldn't introduce positioning since this might cause breaking changes to the layouts of existing slide decks, but might be worth considering for the next major version. This will probably also conflict with some upcoming layout changes I'm working on for the next minor version.

Generally, I'd say avoid using positioning at all costs, but if you want to use absolute positioning within a slide, you can add a custom layout component to wrap the slide with `position: relative`

jesstelford:
Yeah, there's definitely a chance of some breakage.

The original motivator was to break out of the flexbox layout that's applied to the `Slide` component (and is unchangable by a `Layout` component because the heirarchy is `Slide > Layout`), but only for a single element (ie; a logo in the corner)

Am happy to leave this till a later major - I've forked and published `@jesstelford/mdx-deck@1.7.3-patch2` for now which I'll work from.

Thanks!

jxnblk:
Oh, for things like that, you might want to look into using a custom Provider component: https://github.com/jxnblk/mdx-deck/blob/master/docs/advanced.md#custom-provider-component

The docs could use some work, but it lets you wrap all the slides with a component, which is ideal for a logo or twitter handle, etc that stays for the whole presentation

jxnblk:
Also the layout changes I mentioned should hopefully make the per-slide layouts work a little better as well. I'll try to open an issue for that to let people weigh in on

jesstelford:
Thanks for the pointer - I've used a custom Provider to have a sticky footer which works great!

For the logo in particular, the idea was to have it stuck for only one (or two) slides, not the entire deck.

I look forward to the layout changes 👍

jxnblk:
Cool. Thanks for this, but think I'll close this out for now. Hopefully can work something to support your use-case in in the future – and for the Provider, you can also check `props.index` to conditionally render things only for specific slides, if that's a helpful alternative

PR #169, opened 2018-09-17
feat: add --hotPort option

### LINKED ISSUE #168: Allow to pass hot-reload port as option
... otherwise running mdx-deck inside a docker container is impossible

ds82:
This option allows to specify the port for hot reloading

fixes #168

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -32,6 +32,7 @@ const cli = meow(`
 
       -h --host     Dev server host
       -p --port     Dev server port
+      --hotPort     Dev server hot reload port
       --no-open     Prevent from opening in default browser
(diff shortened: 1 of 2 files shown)


cli.js
```diff
@@ -74,6 +75,9 @@ const cli = meow(`
     },
     webpack: {
       type: 'string',
+      hotPort: {
```
jxnblk: This shouldn't be nested, right?

jxnblk: Would love to merge this in, but I think this needs to be resolved

jxnblk:
Closing due to inactivity

PR #172, opened 2018-09-21
Support Page up/down keys

jesstelford:
Particularly useful when using an external clicker such as the Logitech
r400 which sends PgUp/PgDown keys for the left/right arrows.


jxnblk:
Thanks! LGTM

PR #173, opened 2018-09-21
Use single bundle to retain context for withDeck HOC

### LINKED ISSUE #171: Using `withDeck` HOC from installed module throws error during `mdx-deck build`
For a minimal reproduction, see: https://gist.github.com/jesstelford/368df55ea20d239234dd28282b7789ed

In this case, it's `mdx-deck-code-surfer` erroring with:

```
$ mdx-deck build index.mdx
[mdx-deck] building
[mdx-deck] rendering static html
[err] TypeError: Cannot read property 'update' of null
    at new InnerCodeSurfer (node_modules/mdx-deck-code-surfer/dist/deck-code-surfer.js:60:30)
    at processChild (node_modules/react-dom/cjs/react-dom-server.node.development.js:2418:14)
    at resolve (node_modules/react-dom/cjs/react-dom-server.node.development.js:2384:5)
    at ReactDOMServerRenderer.render (node_modules/react-dom/cjs/react-dom-server.node.development.js:2706:22)
    at ReactDOMServerRenderer.read (node_modules/react-dom/cjs/react-dom-server.node.development.js:2680:23)
    at renderToString (node_modules/react-dom/cjs/react-dom-server.node.development.js:3071:25)
    at renderHTML (node_modules/mdx-deck/lib/html.js:65:16)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

jesstelford:
Fixes #171 

The issue was due to the way webpack was configured to bundle `require` calls - it was bundling all the `mdx-deck` code (including `src/context.js` which exports the `withDeck` HOC) into a single file. However, `webpack-node-externals` was used to exclude anything from `node_modules` from the bundle.

This meant if any modules in `node_modules` (such as [`mdx-deck-code-surfer`](https://github.com/pomber/code-surfer)) included `mdx-deck`, it would get a seperate, uncached instance.

ie;

```javascript
// mdx-deck/src/Notes.js
import { withDeck } from './context' // resolves to "bundled-context.js"

// node_modules/mdx-deck-code-surfer/src/deck-code-surfer.js
import { withDeck, updaters } from "mdx-deck"; // resolves to "external-context.js", a completely different instance of the same file.
```

This then means the [`React.createContext()` call](https://github.com/jxnblk/mdx-deck/blob/37999a233ee02efb15619164bc61091e1cc77230/src/context.js#L3) would be made in both `bundled-context.js` _and_ `external-context.js`, resulting in a unique `Provider`/`Consumer` pair for each. But only the `bundled-context.js`'s `<Provider>` would ever have a value set.

Ultimately, this results in the `withDeck` HOC from `external-context.js` setting `deck: null`, resulting in the error we have in #171

cc @pomber

### DIFF
--- a/lib/html.js
+++ b/lib/html.js
@@ -7,7 +7,6 @@ const {
 } = require('react-dom/server')
 const { ServerStyleSheet } = require('styled-components')
 const webpack = require('webpack')
-const nodeExternals = require('webpack-node-externals')
 const rimraf = require('rimraf')
 const mkdirp = require('mkdirp')
 const createConfig = require('./config')
@@ -29,15 +28,6 @@ const getApp = async opts => {
     App: path.join(__dirname, '../dist/entry.js')
   }
   config.target = 'node'
-  config.externals = [
-    nodeExternals({
-      whitelist: [
-        'mdx-deck',
-        'mdx-deck/themes',
-        'mdx-deck/layouts'
-      ]
-    })
-  ]
 
   const compiler = webpack(config)
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -3335,7 +3335,7 @@
     },
     "external-editor": {
       "version": "2.2.0",
-      "resolved": "https://registry.npmjs.org/external-editor/-/external-editor-2.2.0.tgz",
+      "resolved": "http://registry.npmjs.org/external-editor/-/external-editor-2.2.0.tgz",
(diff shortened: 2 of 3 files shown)


jxnblk:
Thanks! I've wanted to get this fixed, but `webpack-node-externals` was added to mitigate issues I was seeing with styled-components SSR – I suspect this will introduce issues when there is a locally installed instance of styled-components, so I'll want to test this out as a prerelease before merging. If you have any time to test this out, it'd be super helpful

jxnblk:
Sorry for the delay and thanks for working on this. I'm going to close this out in favor of changes coming in v2 that should resolve the issue here

PR #179, opened 2018-09-22
Adds Option for Progress or Controls

SaraVieira:
👋

This PR adds the option for removing progress indicator from the bottom or replace it with a bar and also adds the possibility to remove the next and previous buttons.

I also removed babel-preset-stage-o and added the necessary plugins because the stage presets have been deprecated. Src: babeljs.io/blog/2018/07/27/removing-babels-stage-presets

## Why

If you have a full-screen embedded in the slide you click the next and previous buttons by accident instead of what you want to click inside the slide.
About the progress, it's simply a stylistic issue.

## Examples:

No Controls: https://site-xpkeoognew.now.sh/
No Progress: https://site-zexwaztfhl.now.sh/
Bar Progress: https://site-wsmpwppvwv.now.sh/#2


src/Controls.js
```diff
@@ -0,0 +1,43 @@
...
+})
+const Next = styled(Button)([], {
+  position: 'fixed',
+  top: 0,
+  right: 0,
+  bottom: 0,
+})
+
+const Controls = ({ update, theme: { controls } }) =>
+  controls !== false ? (
```
sapegin: Wouldn't it be better to have this condition in the parent?

src/progress/index.js
```diff
@@ -0,0 +1,42 @@
...
+        <Bar
+          index={index}
+          length={length}
+          onClick={index => {
+            update({ index })
+          }}
+        />
+      )
+
+    case 'dots':
```
sapegin:
```jsx
case 'dots':
default:
  return (<Dots ...
```

otherwise you're rendering both dots and bar.

src/Controls.js
SaraVieira: That was my initial idea too but I wasn't sure about having the theme in the provider. Think it's better ?

src/progress/index.js
SaraVieira: You mean I can combine the dots and the default and just return once? 

I didn't know that was valid :/

SaraVieira:
I also removed babel-preset-stage-o and added the necessary plugins because the stage presets have been deprecated.
SRC: https://babeljs.io/blog/2018/07/27/removing-babels-stage-presets

src/progress/index.js
SaraVieira: Fixed :)

asos-albinotonnina:
no mergy mergy? :P

jxnblk:
Thanks for this and sorry I never got around to looking at this/merging it in. For the next major version, I'm considering removing the default controls entirely and making those customizable with `theme.Provider`, so I think I'd rather not introduce the new API to v1. Hope that makes sense and thanks for all your help with mdx-deck!

SaraVieira:
That would be great !

I copied what j made here to my presentations so yay 🎉

Issue #121, opened 2018-08-20
PDF export ignores Appear component


zcei:
Hej,

I've been successfully using `mdx-deck` for my latest presentation, and it was a breeze! 🙏 
Now I wanted to export it as a PDF in order to upload it.

## The Problem

List elements (and potentially everything else) in `<Appear />` is not rendered.

## Expected Behavior

Every component in `<Appear />` with index `n` yields a new page that renders all components up to and including index `n`.

## How to reproduce?

```mdx
import { Appear } from 'mdx-deck'

# I don't see the problem?

<ul>
  <Appear>
    <li>Not</li>
    <li>in</li>
    <li>PDF</li>
  </Appear>
</ul>
---
# The end
```

```sh
mdx-deck pdf pdf-issue.mdx
```

Open `dist/presentation.pdf`. It has only two pages, expected are five pages.

hxuanhung:
Any idea on how to takcle this? I can work on it.

jxnblk:
I suspect there's two primary ways: 

1. An explicit prop passed through context to the Appear component
2. Refactor Appear to use styled-components and a `@media print` rule to control visibility

If you wanna try it out in a PR, I'd be happy to look, but I haven't had time to investigate this myself

hxuanhung:
👍, I will try out in a PR

hxuanhung:
Hi, I've just fired a PR for it here #186. I simply launch Puppeteer with a specigic user-agent and check it in Appear.

zcei:
Done in https://github.com/jxnblk/mdx-deck/pull/186

PR #186, opened 2018-10-01
PDF export ignores Appear component #121 

hxuanhung:


lib/pdf.js
```diff
@@ -4,7 +4,9 @@ const mkdirp = require('mkdirp')
 const puppeteer = require('puppeteer')
 
 module.exports = async (opts = {}) => {
-  const browser = await puppeteer.launch()
+  const browser = await puppeteer.launch({
+    args: ['--user-agent=pdf']
```
zcei: Could we rather do [`page.setUserAgent()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagesetuseragentuseragent) in l. 11?

src/Appear.js
```diff
@@ -29,6 +29,10 @@ export default withDeck(class Appear extends React.Component {
       return children
     }
 
+    if (window.navigator.userAgent === 'pdf') {
```
zcei: I think using a spec compliant user-agent would be great. Maybe smth. like `MDX-Deck/1.7.7 Print/PDF`?

This way later plugins / components have an easier time differentiating if needed. And your check would just be `window.navigator.userAgent.includes('Print/PDF')`

lib/pdf.js
hxuanhung: Yes so we just need to set it for the current page rather the whole instance.

src/Appear.js
hxuanhung: I totally agree.

jxnblk:
Sorry for the delay! This looks great, thanks!

PR #187, opened 2018-10-02
Allow components to consume context in grid mode

zcei:
Hej,

currently `withDeck` sometimes cannot find the proper data, as it's not available in `grid` mode. For the other modes a `<Slide/>` is rendered with `context` being passed in. In `grid` mode this is taken care of by the `<Grid/>` component so it should pass the same data through.


zcei:
I'm sorry for the style changes, did the prettier config change somehow? I did definitely do that on purpose 😆

src/SlideDeck.js
```diff
@@ -49,7 +44,7 @@ export class SlideDeck extends React.Component {
   static defaultProps = {
     slides: [],
     theme: defaultTheme,
-    components: {} ,
```
zcei: Looks like prettier isn't running by default, but my editor settings picked it up and is now formatting code that others contributed before.

jxnblk:
Sorry for the delay and no worries about the prettier changes! This LGTM, thanks!

PR #189, opened 2018-10-03
Correct escape URL

signalwerk:


jxnblk:
Sorry for the delay. This shouldn't be necessary. If you have a use-case where the current implementation does not work, please provide a reproducible example and update the snapshots

signalwerk:
@jxnblk it's too long ago to remember my correct example where it broke. But since the [spec](https://www.w3.org/TR/CSS21/syndata.html#uri) tells us that the URL is valid without the quotation marks you are right. but an image name with `page 2.jpg` is way more likely than `page"2.jpg` I still think my pull request is not bloated with two bytes and then people can use a space in their image-names. That was most likely also my case.

jxnblk:
The space should stay intact in the js string in that example though, shouldn’t it?

signalwerk:
@jxnblk no. it's not working when a space is in the filename.

Issue #183, opened 2018-09-26
PDF export of slides using Split are empty


sylvhama:
I could reproduce with mdx-deck 1.7.1 and 1.7.7.

Create a `deck.mdx` file, copy paste this:
```
import { Split } from 'mdx-deck/layouts'

# LEFT

# RIGHT

export default Split
```

Then run `mdx-deck pdf deck.mdx`. The PDF will be empty however the web version looks fine.

hxuanhung:
it seems similar to #121, I can have a look at it.

orhaneee:
I can confirm that bug.

jxnblk:
I think this should work now in v2. Feel free to reopen if you're still seeing issues

PR #190, opened 2018-10-03
PDF export of slides using Split are empty #183

hxuanhung:


src/Flex.js
```diff
@@ -4,10 +4,7 @@ import { space, width, color } from 'styled-system'
 
 const Flex = styled.div([], {
   display: 'flex',
-  justifyContent: 'center',
-  '@media print': {
-    display: 'none'
-  }
```
jxnblk: Sorry for the delay! This isn't ideal, but this line is what hides the navigation dots during PDF export, so it would be good to handle that somehow with this change

jxnblk:
Going to close this out and keep tracking in #183

PR #197, opened 2018-10-11
Add URL-based alternatives for special modes

awayken:
I ran into an instance where a browser plugin was using the same keyboard shortcut as Presenter mode. Knowing that there are URL-based alternatives to enter these modes might be useful for people.


awayken:
Any thoughts on accepting this pull request?

jxnblk:
Sorry for the delay and thanks for this!

PR #200, opened 2018-10-16
Add flag to disable puppeteer sandbox

jacobmischka:
I understand that this may not be something you'd like to support, but it's useful to me and I figured I might as well submit a pull request.

And apologies for the large diff, prettier did its work it looks like.


jxnblk:
Sorry for the delay and thanks for this! No worries about the prettier changes, this LGTM!

PR #202, opened 2018-10-18
[Appear] "alt" prop to render only one Item at a time

albinotonnina:
# Hello!
Thanks for this tool!

## My problem 
the `Appear` children items would take their space even when invisible. That's obviously the default behaviour of the `visibility` property.

I wanted to render just one Item at a time.

with this PR, `<Appear alt>...</Appear>` will do that.

You could argue that I could have just used a `display:none` rule instead of not rendering at all the invisible elements. I preferred this solution for a couple of reasons, better match of the react lifecycle methods with the user interactions and better control of the UIs through CSS transitions/animations, eg transitions on dom append.

It's a quick PR, if you like the idea I can do some more work on it.

If you wanna test it quickly, I published [@albinotonnina/mdx-deck](https://www.npmjs.com/package/@albinotonnina/mdx-deck)

ps. Sorry for the formatting diffs. There must be something with my configuration and lint-staged on pre-commit. I guess.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mdx-deck",
-  "version": "1.7.7",
+  "version": "1.8.0",
   "description": "MDX-based presentation decks",
   "main": "dist/index.js",
   "bin": {
--- a/src/Appear.js
+++ b/src/Appear.js
@@ -4,42 +4,48 @@ import { withDeck } from './context'
 import { setSteps } from './updaters'
 import { modes } from './constants'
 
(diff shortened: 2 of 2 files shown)


jxnblk:
Thanks! Yeah, there's been some discussion of adding a render prop version of the Appear component, which I think would be a much more flexible way to handle things like this IMO. I'll try to take a look at this over the weekend

jxnblk:
Thanks for the PR, but I think offering a render prop solution will be a little more flexible and am going to close this out

PR #208, opened 2018-10-24
Fix typo in Root PropTypes.

parkerziegler:
Thanks for a great library! This PR fixes a small typo I noticed in `Root`s `PropTypes` while reading the `mdx-deck` source. The pre-commit hook seems to have altered some formatting, but the essential change happens on L27 of the diff (just removing a stray "s").

### DIFF
--- a/src/Root.js
+++ b/src/Root.js
@@ -1,19 +1,20 @@
 import styled from 'styled-components'
-import {
-  width,
-  height,
-  color
-} from 'styled-system'
+import { width, height, color } from 'styled-system'
 
-export const Root = styled.div([], {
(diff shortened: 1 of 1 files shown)


src/Root.js
```diff
@@ -23,12 +24,12 @@ export const Root = styled.div([], {
 Root.propTypes = {
   ...width.propTypes,
   ...height.propTypes,
-  ...color.propTypess
+  ...color.propTypes,
```
parkerziegler: Here's the key change.

jxnblk:
Thanks!

PR #210, opened 2018-10-27
Allow overflow for FullScreenCode

awayken:
In cases where code sample is too large for the slide, presenters can now scroll to show more.


jxnblk:
Ha! Hopefully the code examples aren’t that long most of the time 😄 Thanks!

awayken:
@jxnblk No kidding! Actually, I ran into this when I was presenting on a low resolution screen with a weird aspect ratio. I didn't have much code, but it was getting cut off, and I figured this was a low key fix for that issue.

I hope people don't use this for evil. The last thing an audience wants is to read a hundred lines of code on a tiny slide! 😆

PR #213, opened 2018-11-02
Fix Appear children PropTypes

etienne-dldc:
Hi,

`Appear` children can be a single element so PropTypes should not required an array


jxnblk:
Thanks!

PR #217, opened 2018-11-08
[Appear] keep style of children if exists

Ayc0:
Currenly,`<Appear/>` set its children style props to `{ visibility: ... }`

With this PR, if a child had a pre-existing style, it will be kept

### DIFF
--- a/src/Appear.js
+++ b/src/Appear.js
@@ -35,6 +35,7 @@ export default withDeck(class Appear extends React.Component {
           React.cloneElement(child, {
             key: i,
             style: {
+              ...((child.props || {}).style || {})
               visibility: (step >= i + 1) ? 'visible' : 'hidden'
             }
           })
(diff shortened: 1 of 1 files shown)


src/Appear.js
```diff
@@ -35,6 +35,7 @@ export default withDeck(class Appear extends React.Component {
           React.cloneElement(child, {
             key: i,
             style: {
+              ...((child.props || {}).style || {}),
```
jxnblk: I think you could spread `child.props.style` after the `visibility` property to ensure that the behavior of Appear still works as expected

Ayc0: If you spread the style afterwards and if in the style there is `visibility`, Appear won't do anything

jxnblk: Sorry, guess my early morning brain was thinking about this the wrong way 😬

PR #226, opened 2018-11-18
Upgrade remark-unwrap-images

MicheleBertoli:
Hello : )

I found that copy/pasting markdown links to [CodeSandbox](https://codesandbox.io/) (and, in general, using images wrapped in links) wasn't working as expected, so [I fixed it](https://github.com/johno/remark-unwrap-images/pull/2).

I hope this is OK to merge.

Thank you very much!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9211,11 +9211,11 @@
       }
     },
     "remark-unwrap-images": {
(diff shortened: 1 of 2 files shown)


jxnblk:
Looks great! Thanks!

Issue #231, opened 2018-11-29
Support for .jsx extension


jollytoad:
I've just attempted to add my first component into an mdx-deck presentation, and hit this issue which stumped me for a while... I was getting:

`Module not found: Error: Can't resolve './components/Foo' in '...'`

after much head scratching, and searching, i realised my component ended in .jsx - as usual, but mdx-deck doesn't support this. Renaming to .js works - but breaks my editor.

Any chance support for .jsx could be added?

Thanks
- Mark

jxnblk:
This should be supported now, so closing this out

PR #237, opened 2018-12-14
Support .jsx components

leonrodenburg:
This adds support for importing components with a `.jsx` file extension. Webpack will also resolve these if no extension is provided.

For example:
`import FancyComponent from './components/Fancy'` 
(will resolve to either `Fancy.js` and `Fancy.jsx`)

Solves #231

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -7,7 +7,7 @@ const merge = require('webpack-merge')
 const chalk = require('chalk')
 const remark = {
   emoji: require('remark-emoji'),
-  unwrapImages: require('remark-unwrap-images')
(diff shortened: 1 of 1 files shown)


lib/config.js
```diff
@@ -16,54 +16,47 @@ const babel = {
...
-    test: /\.js$/,
+    test: /\.jsx?$/,
     exclude: /node_modules/,
     loader: require.resolve('babel-loader'),
-    options: babel
+    options: babel,
   },
   {
-    test: /\.js$/,
+    test: /\.jsx?$/,
```
jxnblk: This package doesn't use the `.jsx` extension, so this change shouldn't be necessary

PR #239, opened 2018-12-17
Support .jsx components

leonrodenburg:
This adds support for importing components with a .jsx file extension. Webpack will also resolve these if no extension is provided.

For example:
import FancyComponent from './components/Fancy'
(will resolve to either Fancy.js and Fancy.jsx)

Solves #231
Cleaned up version of #237


jxnblk:
Thanks for this! I only use `.js` file extensions at this point, but no harm in supporting the `.jsx` file extension as far as I can tell

Issue #157, opened 2018-09-09
puppeteer as peerdependency


timneutkens:
Not everyone might have the use case of exporting to pdf / creating a screenshot. In my case I'm on a 500Kbps connection at home and installing mdx-deck via simple `yarn add mdx-deck` takes so long that I had to switch to my phone 4G connection which is 40Mbps. Even then it takes 50seconds to install though. So the simplest solution to me seems like making puppeteer a peer-dependency 🙏

jxnblk:
Yeah, been pondering this. I wonder if the PDF script should be a separate package altogether, rather than in the main CLI. Something like `mdx-deck-pdf`

timneutkens:
Sounds like a great solution 👍:pray:

millette:
I didn't want to mention it since I saw the pdf export was issue #3 (so, part of the original plan). Glad you're ok with it :-)

simonhaenisch:
Another approach I know (from [StencilJS](https://github.com/ionic-team/stencil/blob/master/src/sys/node/node-lazy-require.ts)) is "lazy dependencies":

```js
// package.json
{
  "lazyDependencies": {
    "puppeteer": "^1.10.0"
  }
}
```

Then, only when the `pdf` sub-command is invoked, run a function that ensures that the packages required for that command are installed, and install them if missing.

Siilwyn:
Another solution would be using `puppeteer-core` this is an easy switch, the only difference being it expects chrome to be installed.

jxnblk:
Removed puppeteer from the core CLI in v2, so hopefully it'll be a smaller install size going forward – PDF/PNG export is now available with the separate `@mdx-deck/export` CLI package

styfle:
|version | install size |
|-------:|--------------|
| 1.8.2 | [![install size](https://packagephobia.now.sh/badge?p=mdx-deck@1.8.2)](https://packagephobia.now.sh/result?p=mdx-deck@1.8.2) |
| 2.0.0-8 | [![install size](https://packagephobia.now.sh/badge?p=mdx-deck@2.0.0-8)](https://packagephobia.now.sh/result?p=mdx-deck@2.0.0-8) |

Looks good! 🎉

PR #240, opened 2018-12-18
Add badge to display install size

styfle:
This adds a badge to the `README.md` to display the npm install size

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,12 +7,15 @@
 [![Build Status][badge]][travis]
 [![Version][version-badge]][npm]
 [![Downloads][downloads-badge]][npm]
+[![Install Size][install-size-badge]][install-size]
 
 [badge]: https://img.shields.io/travis/jxnblk/mdx-deck.svg?style=flat-square
 [travis]: https://travis-ci.org/jxnblk/mdx-deck
 [version-badge]: https://img.shields.io/npm/v/mdx-deck.svg?style=flat-square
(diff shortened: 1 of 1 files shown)


jxnblk:
Since this is a dev/build tool, it's not really intended to be installed in other applications. I don't think it makes sense to advertise that this is a (necessarily) massive CLI install

styfle:
There is at least one other person who cares about the install size of their tools: https://github.com/styfle/packagephobia/issues/218

I realize it's not a goal but I'm trying to bring awareness because I don't believe there is any other way for users to know the size of `mdx-deck` before installing.

Siilwyn:
Yes! I like my dev/build tools lightweight too, the 360MB install size is definitely a bummer and good to know beforehand.

Siilwyn:
Btw. this is very much related to: #157

jxnblk:
v2 has removed the `puppeteer` dependency from the core CLI, which should substantially reduce the install size. I took another look at this PR and it seems like the image link is either broken or *extremely* slow. I'm going to close this out for now

styfle:
The badge is probably timing out when installing all the dependencies.

|version | install size |
|-------:|--------------|
| 1.8.2 | [install size](https://packagephobia.now.sh/result?p=mdx-deck@1.8.2) |
| 2.0.0-8 | [install size](https://packagephobia.now.sh/result?p=mdx-deck@2.0.0-8) |

Once you publish v2 as `latest` to npm, it will work nicely 👍

PR #246, opened 2019-01-20
Add mdx-deck-live-code as external library

JReinhold:
Thanks for mdx-deck, it is really awesome!

I made this library to do live React and JS coding in the slides. I would be honoured to be mentioned in the Readme.
I would also really appreciate some feedback from anyone who has time someday.

repo: https://github.com/JReinhold/mdx-deck-live-code
demo: https://reinhold.is/live-coding-in-slides


jxnblk:
Looks awesome, thanks!

PR #250, opened 2019-01-30
Fix default language for syntax highlighting

rexxars:
I apologize for the big change here, it seems that prettier has not been run on these files, but is now being applied in a precommit hook?

Anyway:

This fixes a tiny little bug where the default language in the `<Code />` component uses the string `javascripts` instead of `javascript`. It also fixes a typo: `getLangauge` => `getLanguage`.


jxnblk:
No worries about the prettier changes. Thanks for this and LGTM!

PR #251, opened 2019-01-30
Include prism styling by default

rexxars:
This is a bit opinionated, I realize, but it seems weird to me that there is no syntax highlighting styles included by default. It wasn't super clear to me how to add it, and after finding the solution, I feel like it is perhaps a bit more work than necessary:

- Add a `theme.js`
- Figure out the path and import of the default theme
- Figure out the path and import a prism styling config
- Spread the default styles onto the export
- Add an export declaration for the theme in the deck

Since the actual syntax highlighting for JS/JSX is being included by default, it seems a bit of an add odd choice not to also include a theme so that it "just works", but can still be configured.

I investigated how many extra bytes it would add to include the `okaidia` theme:
```
Approximate weight of react-syntax-highlighter/dist/styles/prism/okaidia:
  Uncompressed: 3.91 kB
  Minified (uglify): 2.47 kB
  Minified and gzipped (level: default): 973 B
```

Which doesn't seem like much to me, compared to the convenience it gives.

I don't have any strong feelings on which theme should be used, I just picked one I personally found pleasing. The other changes in this PR is due to prettier being applied in a precommit stage.

### DIFF
--- a/src/themes/base.js
+++ b/src/themes/base.js
@@ -1,9 +1,9 @@
+import okaidia from 'react-syntax-highlighter/dist/styles/prism/okaidia'
+
 export default {
   font: 'system-ui, sans-serif',
   monospace: 'Menlo, monospace',
-  fontSizes: [
-    '0.75em', '1em', '1.5em', '2em', '3em'
-  ],
+  fontSizes: ['0.75em', '1em', '1.5em', '2em', '3em'],
   colors: {
     text: '#000',
     background: 'white',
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks for the PR, but I think I'd rather keep syntax highlighting decoupled from the defaults. There's a strong case for *not* wanting syntax highlighting in some themes, and, with recent changes in react-syntax-highlighter's API, I don't want to couple mdx-deck too closely together, if that makes sense.

Hopefully in the next version, it'll be a little bit clearer on how to add your own syntax highlighting components

PR #253, opened 2019-02-06
Prevent Appear from resetting when changing slides

TrevorBurnham:
Howdy! First-time contributor here.

Currently, `Appear` resets to the first step when you change slides, causing its content to disappear during the transition to the next slide. You can see this [in the demo deck](https://jxnblk.com/mdx-deck/#9.4):

appear-glitch

This is caused by `Appear` receiving a new `deck` object with `step={0}` when the transition occurs. The fix in this PR is to have `Appear` ignore new `step` values when the `deck` it receives has `active={false}`.

Please let me know if there's anything else you need from me here.

### DIFF
--- a/src/Appear.js
+++ b/src/Appear.js
@@ -11,10 +11,21 @@ export default withDeck(
       deck: PropTypes.object.isRequired,
     }
 
+    static getDerivedStateFromProps(props) {
+      const { deck } = props
+      if (!deck.active) return null
+      return {
+        step: deck.step,
+      }
+    }
+
     constructor(props) {
       super(props)
       const { update, index } = props.deck
       const steps = React.Children.toArray(props.children).length
+      this.state = {
+        step: 0,
+      }
       update(setSteps(index, steps))
     }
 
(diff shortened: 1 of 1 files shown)


jxnblk:
Good catch! Looks like there's some conflicts, if you have time to sort those out. Would be great if this didn't rely on `getDerivedStateFromProps`, but would need to look at it a little more closely to provide more direction

TrevorBurnham:
@jxnblk Sure thing. I've rebased against `master` to resolve the conflicts.

> Would be great if this didn't rely on `getDerivedStateFromProps`

It needs some kind of lifecycle method to update state, either `getDerivedStateFromProps` or `componentWillReceiveProps`. And it looks like this project is all-in on React 16: You upgraded to styled-components v4 as part of #259. So I don't see what benefit you'd get from switching from `getDerivatedStateFromProps` to another lifecycle method. Or did you have something else in mind?

jxnblk:
Sorry for the delay – going to merge this in as-is since it's certainly better than the current behavior. Thanks for this!

PR #255, opened 2019-02-10
feat(webpack): enable user customize webpack externals whitelist to f…

### LINKED ISSUE #254: enable customize node externals config
right now the config is fixed here https://github.com/jxnblk/mdx-deck/blob/master/lib/html.js#L32

to fix code-surfer issue when building https://github.com/pomber/code-surfer/issues/8#issuecomment-422095168, we have 2 options:

whitelist mdx-code-surfer there, 
or enable user manage this

what do you think?

sibelius:
…ix mdx-code-surfer, fix #254

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -4,10 +4,11 @@ const webpack = require('webpack')
 const HTMLPlugin = require('mini-html-webpack-plugin')
 const ProgressBarPlugin = require('progress-bar-webpack-plugin')
 const merge = require('webpack-merge')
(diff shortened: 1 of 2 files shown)


jxnblk:
Thanks! Hopefully in the next version, the whitelist/node externals stuff can be ditched, but if this fixes some issues, it seems like a good temporary solution

sslotsky:
@jxnblk looks like this did not go into 1.10.0, any idea when this will be released? :pray:

PR #261, opened 2019-02-20
Adds documentation for undocumented `withSlide`

chrisjpatty:
I had a use-case where I needed access to the "active" state of the current slide so I could reset the animation in a custom component. After digging around in the source I discovered the undocumented `withDeck` and `withSlide` functions. I couldn't distinguish any difference between the two when I tried them, but the `withSlide` function was exactly what I needed.

This pull request just adds a line to `advanced.md` describing how to use `withSlide` to get access to the current active state of the slide.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -49,6 +49,26 @@ which can be used to show custom page numbers or add other elements to the UI.
 - `notes`: (object) custom [speaker notes](#speaker-notes) for all slides
 - `step`: (number) the current visible step in an Appear component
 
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks for the PR! I'd rather keep this API somewhat "private" for v1 since the changes coming in v2 are going to handle the context and internal state a bit differently – you can certainly use `withSlide` for v1 but v2 should have some nicer mechanisms for picking up state from mdx-deck

PR #274, opened 2019-03-13
Now 2.0 deployment ready

paulogdm:
I just did a "copy & paste" of the previous `now.json` and also added a `.nowignore`.

### DIFF
--- a/now.json
+++ b/now.json
@@ -1,0 +1,21 @@
+{
+  "version": 2,
+  "public": true,
+  "name": "mdx-deck",
+  "alias": "mdx-deck.jxnblk.com",
+  "builds": [
+    {
+      "src": "docs/package.json",
+      "use": "@now/static-build",
+      "config": {
+        "distDir": "dist"
+      }
+    }
+  ],
+  "routes": [
+    {
+      "src": "/(.*)",
+      "dest": "/docs/$1"
+    }
+  ]
+}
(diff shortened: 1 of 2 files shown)


now.json
```diff
@@ -0,0 +1,25 @@
...
+  ],
+  "routes": [
+    {
+      "src": "/main.js",
+      "dest": "/docs/main.js"
+    },
+    {
+      "src": "/(.*)",
+      "dest": "/docs/index.html"
+    }
```
jxnblk: Not that I plan to immediately, but this wouldn't pick up things like static assets, would it? I wonder if this last route can only do `.html` requests?

paulogdm: It seems you are correct.
I think we may have a bug on our "Output" of "/_src": 
image

paulogdm: I mean, does `mdx` uses any static assets after the build?

paulogdm: Yes, you are correct. But we can add routes later for that case. Something like:
```
    {
      "src": "/static/(.*)",
      "dest": "/docs/static/$1"
    }
```
But the fallback for `/docs/index.html` is still critical for a SPA.
https://github.com/zeit/now-examples/blob/master/create-react-app/now.json

jxnblk:
Thanks for the help with this! I think for now I'm going to keep it on netlify – already have DNS/etc set up, but I'm sure I'll use this PR for future reference when I have to figure out how to get now set up

PR #290, opened 2019-03-23
Take loaders in account when merging webpack configs

kiurchv:
This change adds the ability to redefine configurations of webpack loaders that are set by internal config.

For example, to use additional remark plugins they should be passed to mdx-loader. Currently, it is not possible to manually configure mdx-loader since it always set by internal webpack config and if we add it into own config then two instances of mdx-loader will be configured.


jxnblk:
Thanks!

PR #293, opened 2019-03-25
Show all Appear steps in presentation mode next slide preview

### LINKED ISSUE #76: Show all Appear steps in presentation mode next slide preview
The next slide preview should show all steps revealed to be useful.

vlsi:
fixes #76

### DIFF
--- a/packages/components/src/Presenter.js
+++ b/packages/components/src/Presenter.js
@@ -6,10 +6,13 @@ import Pre from './Pre'
 import Clock from './Clock'
 
 export const Presenter = props => {
-  const { slides, index } = props
+  const { slides, index, step } = props
   const Current = slides[index]
-  const Next = slides[index + 1]
-  const { notes } = Current.meta || {}
+  const { notes, steps } = Current.meta || {}
+  // "next" slide in Presenter includes all the animations
+  const nextStep = step < steps ? step + 1 : 0
+  // nextStep==0 means we are out of steps and/or the current slide has no steps
+  const Next = nextStep == 0 ? slides[index + 1] : Current
 
   return (
     <div
@@ -48,7 +51,7 @@ export const Presenter = props => {
         >
           <Zoom zoom={1 / 4}>
             {Next && (
-              <Slide>
+              <Slide step={nextStep}>
                 <Next />
               </Slide>
             )}


Taar:
I'm wondering if maybe this should be handled by the components in each slide rather than setting the `step` to the last step for the given slide. The `Slide` component should also be getting the current `mode` ( see my PR #299 ) and since the component needs to register how many steps it needs, it already has that information. So if the `mode` is print, the component will show all the steps. I just feel there might be a case where you wouldn't want to show all the steps or maybe not the last step? But I suppose, if the `mode` is print, the component could also ignore the current `step` altogether. Either way, I think it's important for the the `Slide` component to get the `mode` too. Thoughts?

vlsi:
> rather than setting the step to the last step for the given slide

What do you mean by that?

Taar:
Sorry let me clarify, I think using the `mode` instead of the `steps` to show all the step elements in `presenter` mode would be better. Currently, the `Slide` component should also be passed a prop of `mode` which will be used to populate the context provider [(See here)](https://github.com/jxnblk/mdx-deck/blob/master/packages/components/src/Slide.js#L20). So `Appear` could use that to know when to show all the step elements. Therefore, each component in the slide can decided based on the mode what they should show or hide. As of right now, `Appear` uses `useSteps` to get the current steps, so maybe there should also be another hook called, `useModes` which would return the current `mode`. With that `Appear` could choose to show all its step elements in the `presenter` mode.

vlsi:
> With that Appear could choose to show all its step elements in the presenter mode.

I'm afraid you don't get the point.
One of the most important features of "presenter" mode is to see *next* slide.

I might want to see "2-3-5 next" slides, however the most important is to know what the audience will see when I click just once.

I don't need `the last step for the given slide`. I don't need `all the step elements in presenter mode`.

I need *next* one which is either subsequent slide or the current one with step increased by one.

Taar:
Ah, yes, I did misunderstand, sorry about that. Though, I think it would be a good idea to also pass `Slide` all the other props like `mode`, `index`, etc.

vlsi:
> Though, I think it would be a good idea to also pass `mode`, `index`, etc

What for?

`index` is already available in `mdx-deck 2`:
https://github.com/pomber/code-surfer/pull/54/files#diff-a170f727a3e688efdc69851fa03f997dR20

I'm not sure if `mode` is available in the same way or not.

Taar:
It's more of a nice to have. #299 was meant to fix this issue. The reason I brought it up was because `Slide` sets the context for all its children based on the props it gets. If instead you used the `Steps` component, `index` would be `undefined` since it's not set on the closest provider.

vlsi:
>If instead you used the Steps component, index would be undefined since it's not set on the closest provider

Well, in my current presentation I'm using both `<Steps`, `<Appear`, and the fix seem to be working for `presenter` mode.

jxnblk:
Thanks for looking into this. I have some code to handle this in #323 (among some other updates), setting a default prop on the `Slide` component here: https://github.com/jxnblk/mdx-deck/pull/323/files#diff-0bc2b66083bf9af93cb7c1617788c4c0R18 should work so that when context isn't available to the component, it will show all steps in the Appear

vlsi:
@jxnblk , I don't see how #323 fixes presenter mode.

Here's how current mdx-deck HEAD looks like: 
Снимок экрана 2019-04-14 в 0 13 53

The slide is as follows (~demo.mdx)

```
### Appear

<ul>
  <Appear>
    <li>One</li>
    <li>Two</li>
    <li>Three</li>
    <li>Four</li>
  </Appear>
</ul>
```

So "preview" should show `one two three`, however it skips `Appear` and it shows the subsequent slide which is wrong.

jxnblk:
The issue #76 only suggests making the steps appear in the next slide not the current slide, which I think makes sense

jxnblk:
Thanks for looking into this but going to close this one out for now

PR #295, opened 2019-03-26
build: add non-ssr-build

pengx17:
Build is building in server render mode by default, but it eliminates some possibilities such as usage of `document` web api. Some users (like me) does not always have a need to serve the page from a node server.


jxnblk:
Thanks! There was a `--no-html` flag in v1 and if you add this, I think the API should stick to the previous one

jxnblk:
Thanks for this!

PR #296, opened 2019-03-26
Update Advanced Docs

Taar:
Provide an example of how to define your own custom Provider and clarify how to set a Provider in your custom theme. Hopefully this will save folks a bit of time when they are looking into creating their own custom Provider.

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -20,8 +20,9 @@ at the [default set of components](../packages/components/src/mdx-components.js)
 
 ## Custom Provider component
 
-A custom Provider component can be added to the theme to wrap the entire application.
-This is useful for adding custom context providers in React or adding persistent UI elements to the entire deck.
+A custom Provider component is useful for adding custom context providers in React or adding persistent UI elements to the entire deck.
+
+To define a custom Provider component, you'll need to import it into your custom theme and set it using the key `Provider` like shown below: 
 
 ```js
 // example theme.js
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks for this! Have a few comments, if you don't have time to work on this, I can try to make the edits over the weekend

docs/advanced.md
```diff
@@ -44,6 +45,48 @@ which can be used to show custom page numbers or add other elements to the UI.
...
 
+#### Example
+
+The example below will display the current slide out of the total amount of slides.
+
+```js
+import React, { useEffect, useRef, useState } from 'react'
+import { css } from '@emotion/core'
+
+function AtTheBottomCenter ({ children }) {
```
jxnblk: I think this whole example would be simpler to read without the additional window logic that could be handled with CSS.

docs/advanced.md
```diff
@@ -44,6 +45,48 @@ which can be used to show custom page numbers or add other elements to the UI.
...
+    const width = ref.current.clientWidth
+
+    setLeft((bodyWidth / 2) - (width / 2))
+  })
+
+  const css = css`
+    position: absolute;
+    bottom: 0;
+    color: #ffffff;
+  `
```
jxnblk: Thought you could do multiline suggested changes, but I can't figure it out... I'd prefer this inline in the component below, but the CSS could be:

```js
const css = {
  position: 'absolute',
  bottom: 0,
  left: '50%',
  transform: 'translateX(-50%)',
}
```

Or, it could span the entire width and center the text:
```js
const css = {
  position: 'absolute',
  left: 0,
  right: 0,
  bottom: 0,
  textAlign: 'center',
}
```

docs/advanced.md
Taar: Ah, yes I do agree it would be simpler if I removed that bit of logic. I should be able to make the changes at some point this weekend. 😄

docs/advanced.md
Taar: Yes I agree, using css here would be a lot better.

docs/advanced.md
```diff
@@ -44,6 +45,37 @@ which can be used to show custom page numbers or add other elements to the UI.
 - `step`: (number) the current visible step in an Appear or Step component
 - Each slide includes a `meta` object with a `notes` field when the Notes component is used within a slide
 
+#### Example
+
+The example below will display the current slide out of the total amount of slides.
+
+```js
+import React from 'react'
+import { css } from '@emotion/core'
```
jxnblk:
```suggestion
```

docs/advanced.md
```diff
@@ -44,6 +45,37 @@ which can be used to show custom page numbers or add other elements to the UI.
...
+    position: 'absolute',
+    left: 0,
+    right: 0,
+    bottom: 0,
+    color: #ffffff;
+    textAlign: 'center',
+  }
+
+  return <div css={ css }>
+    { children }
```
jxnblk: Seems like prettier isn't formatting this as expected. I'll clean up the syntax after merging

docs/advanced.md
```diff
@@ -44,6 +45,37 @@ which can be used to show custom page numbers or add other elements to the UI.
 - `step`: (number) the current visible step in an Appear or Step component
 - Each slide includes a `meta` object with a `notes` field when the Notes component is used within a slide
 
+#### Example
+
+The example below will display the current slide out of the total amount of slides.
+
+```js
+import React from 'react'
```
jxnblk:
```suggestion
// example Provider.js
import React from 'react'
```

jxnblk:
Sorry for the delay on this, and thanks again!

PR #299, opened 2019-03-28
Grid/Overview/Presenter props fix

### LINKED ISSUE #297: Grid doesn't pass correct props to Slide component
`Slide` uses the props passed as the object for its `Context.Provider`. When the program navigates to grid mode, the `Grid` component does not supply additional props. Therefore, `props` is `undefined` which means any component that uses the context will always get `undefined`. 

In the case of `Appear` it seems to work fine because `i < step` will return `false` if `step` is `undefined`.

I think the best way to fix this would be to define default prop values for `Slide` or perhaps using `Proptypes` to warn if the correct props aren't passed in. In that case, `Grid` would need to be updated to pass in the correct props with default values.

Let me know what you think. I could probably have a PR for this in the next day or so if you'd like :smile:

I would also like to update the Advanced docs with some examples on how to use `useSteps` and `Steps`.

**Update:**

`Overview` and `Presenter` also don't pass any props to `Slide` so they'll need to be updated as well.

If anyone runs into this, `step` is `undefined` when using `useSteps`, you can get around this by doing something like below which defaults to the first step:

```js
export default function MyStepper () {
  const step = useSteps(5)
  const currentStep = step != null ? step : 0 // 0 being the first step
  // ...
}
```

Taar:
- Update Slide to include PropTypes to make sure the correct props are
passed
- Move modes to their file so that they can be referenced from another
file
- Update docs demo to include a component that uses the `useSteps`
hooks. This should catch any regression until tests can be created
- Update Grid, Presenter, and Overview to correctly pass in the required
props to Slide defined by its PropTypes

fixes #297

### DIFF
--- a/docs/CodeStepper.js
+++ b/docs/CodeStepper.js
@@ -1,0 +1,38 @@
+import React from 'react'
+import { Prism as SyntaxHighlighter } from 'react-syntax-highlighter'
+import { atomDark as codeStyle } from 'react-syntax-highlighter/dist/esm/styles/prism'
+import { css } from '@emotion/core'
+
+import { useSteps } from '@mdx-deck/components'
+
+const CodeStepper = ({ code, steps, title = '', language = '' }) => {
+  const step = useSteps(steps.length - 1)
+  const current = steps[step]
+
+  return (
+    <>
+      <span
+        css={css`
+          font-size: 2rem;
+          color: #bbbbbb;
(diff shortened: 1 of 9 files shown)


jxnblk:
Sorry I never got around to this PR, and thanks for working on it! Going to close this out in favor of the refactor in #392

PR #300, opened 2019-03-28
Update useSteps to use useEffect

Taar:
- Update useSteps to use useEfect instead of useMemo.
- Rework useSteps.js to allow for testing by passing in a Context into a
factory function
- Add tests for useSteps

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -21,6 +21,8 @@
   "devDependencies": {
     "react": "^16.8.3",
     "react-dom": "^16.8.3",
-    "react-test-renderer": "^16.8.4"
+    "react-hooks-testing-library": "^0.3.8",
+    "react-test-renderer": "^16.8.4",
+    "sinon": "^7.3.1"
   }
 }
--- a/packages/components/src/useSteps.js
+++ b/packages/components/src/useSteps.js
@@ -1,13 +1,16 @@
-import { useContext, useMemo } from 'react'
-import { Context } from './context'
+import { useContext, useEffect } from 'react'
+import { Context as C } from './context'
 
-export default length => {
+export const useStepsFactory = Context => length => {
   const context = useContext(Context)
   const { register, index, step } = context
-  useMemo(() => {
+
+  useEffect(() => {
     if (typeof register !== 'function') return
     register(index, { steps: length })
   }, [length])
 
   return step
 }
+
+export default useStepsFactory(C)
--- a/packages/components/src/__tests__/useSteps.js
+++ b/packages/components/src/__tests__/useSteps.js
@@ -1,4 +1,4 @@
-import React, { createContext } from 'react'
+import { createContext } from 'react'
 import sinon from 'sinon'
 import { renderHook, cleanup } from 'react-hooks-testing-library'
 import { useStepsFactory } from '../useSteps'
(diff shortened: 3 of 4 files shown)


jxnblk:
Thanks for this! Might need a little more context into the motivation for this changes...

packages/components/package.json
```diff
@@ -21,6 +21,8 @@
   "devDependencies": {
     "react": "^16.8.3",
     "react-dom": "^16.8.3",
-    "react-test-renderer": "^16.8.4"
+    "react-hooks-testing-library": "^0.3.8",
+    "react-test-renderer": "^16.8.4",
+    "sinon": "^7.3.1"
```
jxnblk: I'd rather not add the extra dependencies – I'm not sure if `sinon` adds anything that Jest doesn't already have and for testing hooks, I'd rather see unit tests for components that use the hooks

packages/components/src/useSteps.js
```diff
@@ -1,13 +1,16 @@
-import { useContext, useMemo } from 'react'
-import { Context } from './context'
+import { useContext, useEffect } from 'react'
+import { Context as C } from './context'
 
-export default length => {
+export const useStepsFactory = Context => length => {
```
jxnblk: I think the mdx-deck context should be the only context used here – feel free to copy/fork the code if you want this functionality elsewhere

packages/components/src/useSteps.js
```diff
@@ -1,13 +1,16 @@
...
   const context = useContext(Context)
   const { register, index, step } = context
-  useMemo(() => {
+
+  useEffect(() => {
```
jxnblk: I think I get the rationale here, but the reason I went with memo was that it *might* work server-side – this is definitely a "side effect" and it might not matter for static rendering

packages/components/src/useSteps.js
Taar: This was meant to make the hook testable by its self and without having to mock out the context. This way you can pass in a context that's contained in the test.

packages/components/package.json
Taar: My bad, I'm not familiar with Jest and I should have looked to see if it had spies built in. I do agree with testing components that use the hooks however, I feel that there also should be tests specifically testing the hook. That way it's a bit more explicit to what the hook is expected to do. Also if in the future a test needs to be created to make sure there isn't any regression.

packages/components/src/useSteps.js
Taar: I'd have to do some additional research into how the `useEffect` hook works when rendering on the server. I believe it should work no differently then in the browser, especially since it's not using anything browser specific? I'll circle back once I've looked into it.

jxnblk: Revisiting this, I think `useEffect` is indeed the one to use for this sort of thing 👍

packages/components/package.json
jxnblk: The thinking with not testing the hooks directly is that I'd rather test the end API for what the hook does. It's unlikely, but if the implementation changes, I'd rather the unit tests still pass if the API still works, if that makes sense

jxnblk:
Heads up, I'm doing some major refactoring to use react hooks, so you may get some conflicts soon. If you'd like to keep working on adding some unit tests, I can give some pointers once things settle a little more

Taar:
You can close this PR if that's the case. 

Sure thing, I wouldn't mind adding more unit tests and even looking into functional tests as well.

Issue #303, opened 2019-03-28
Build fails when using syntax-highlighter-prism


Kampfheizung:
Hey guys,

currently we're trying to finish our 2.0 migration but at the finish line we encountered 2 build errors that hinders us to use the new fancy major version.

1 error inside `./node_modules/mdx-deck/node_modules/@mdx-deck/themes/src/syntax-highlighter-prism.js`
```js
Module parse failed: Unexpected token (11:9)
You may need an appropriate loader to handle this file type.
| export const createCode = (opts = {}) => props => {
|   const language = getLanguage(props.className)
>   return <Prism {...opts} language={language} {...props} />
| }
| 
```

and 1 error inside `./node_modules/mdx-deck/node_modules/@mdx-deck/themes/src/syntax-highlighter.js`

```js
Module parse failed: Unexpected token (17:9)
You may need an appropriate loader to handle this file type.
| export const code = props => {
|   const language = getLanguage(props.className)
>   return <SyntaxHighlighter language={language} {...props} />
| }
|
```

our theme.js is looking like that:

```js
import theme, {syntaxHighlighterPrism} from '@mdx-deck/themes'
import okaidia from 'react-syntax-highlighter/dist/esm/styles/prism/okaidia'

export default syntaxHighlighterPrism({
  ...theme,

  css: {
    ...theme.css,
    backgroundImage: 'url(../images/background.svg)',
    backgroundPosition: 'top right',
    backgroundRepeat: 'no-repeat'
  },

  colors: {
    ...theme.colors,
    text: 'rgb(61, 77, 101)'
  },

  prism: {
    style: okaidia
  }
});

```

Do you guys have any suggestion how to proceed on that? Or what the problem really is?

Best regards,
the @farbenmeer team

jxnblk:
Sorry for the delay! This looks like a babel issue. If you can point to a reproduction of this issue that'd be super helpful. Also could be helpful to see the full list of dependencies in a gist

jxnblk:
Feel free to reopen if you can provide a reproduction

PR #308, opened 2019-04-02
Update syntax highlighting docs with an example

joshprice:
Couldn't get this to work until I saw the example in #303 which finally made this work.

### DIFF
--- a/docs/theming.md
+++ b/docs/theming.md
@@ -71,6 +71,15 @@ Themes can provide a set of custom MDX components, including a replacement for t
 
 MDX Deck includes two themes for adding syntax highlighting with [react-syntax-highlighter][]: `syntaxHighlighter` and `syntaxHighlighterPrism`.
 
+```js
+import { future } from "mdx-deck/themes";
+import { syntaxHighlighterPrism } from "mdx-deck/themes";
+
+export default syntaxHighlighterPrism({
+  ...future
+});
+```
+
 Since MDX supports using React components inline, you can also import a syntax highlighting component directly, if you prefer.
 
 ### Styling Elements


docs/theming.md
```diff
@@ -71,6 +71,15 @@ Themes can provide a set of custom MDX components, including a replacement for t
 
 MDX Deck includes two themes for adding syntax highlighting with [react-syntax-highlighter][]: `syntaxHighlighter` and `syntaxHighlighterPrism`.
 
+```js
+import { future } from "mdx-deck/themes";
+import { syntaxHighlighterPrism } from "mdx-deck/themes";
+
+export default syntaxHighlighterPrism({
+  ...future
+});
```
jxnblk: Thanks for putting this together! I think the preferred API should be in the section above this: https://github.com/jxnblk/mdx-deck/blob/master/docs/theming.md#composing-themes

joshprice: There is no description at all currently of how this works. This example would have definitely saved me some time and frustration at the time. So agree that an even better description of the API above would be lovely. Don't think I'm the person to write it though (although I'd love to help), mainly because it's unclear how it does actually work...

joshprice:
Are they meant to be doing the same thing? If so then that wasn’t clear to
me.

This was a while ago now, but I absolutely tried the composition with
arrays first and that did not work - I got no syntax highlighting. Hence my
confusion.

I can attempt again as it’s possible it didn’t work for some other reason.

jxnblk:
> This was a while ago now, but I absolutely tried the composition with
arrays first and that did not work - I got no syntax highlighting.

^^ yeah so this sounds like a bug – could be related to #284

jxnblk:
Going to close this for now since the example added here is *not* intended to be part of the public API

Issue #310, opened 2019-04-04
Switch slides on PageUp/Down keys


vlsi:
PageUp/Down is commonly used by clickers/presenters, so it makes sense to support that by default.

unitehenry:
For these clickers, does the page up advance the slide or go back?

vlsi:
For left-to-right languages pagedown is to advance the slides.
I've no idea what's the expected operation for right-to-left or top-to-bottom languages.

vlsi:
Thanks for mdx-deck, it looks like the project is not suitable for presentations on stage: https://github.com/jxnblk/mdx-deck/pull/293

PR #319, opened 2019-04-12
Added pageUp and pageDown keys

unitehenry:
Added the ability to navigate slides with pageUp and pageDown as requested in issue #310


packages/components/src/MDXDeck.js
```diff
@@ -90,10 +93,12 @@ export class MDXDeck extends React.Component {
       }
     } else {
       switch (keyCode) {
+        case keys.pgDown:
```
jxnblk: Thanks for this! I might be totally wrong, but shouldn't `pageDown` advance to the next slide?

unitehenry: Good question. I'll ask the author of the issue.

jxnblk: Based on what Google Slides does it seems to be page down for next slide

packages/components/src/MDXDeck.js
```diff
@@ -35,6 +35,8 @@ const keys = {
   p: 80,
   o: 79,
   g: 71,
+  pgUp: 33,
```
vlsi: By the way, key codes are exactly the ones I've used during my presentation

unitehenry:
Ok pushed the changes.

jxnblk:
Thanks!

PR #323, opened 2019-04-12
Slide embed component

jxnblk:


packages/components/src/Ratio.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+export const Ratio = ({ ratio, children }) => (
+  <div
+    css={{
+      width: '100%',
+      height: 0,
+      paddingBottom: ratio * 100 + '%',
```
vlsi: This sounds like a funny hack, however one would still need a way for "absolute positioning within the slide".

For instance:
1) Slide numbers are typically placed among the corners
2) Overlay captions

packages/gatsby-theme/src/layouts/index.js
```diff
@@ -44,26 +30,33 @@ const Thumb = props => (
...
+      {props.decks.map(deck => (
+        <li
+          key={deck.id}
+          css={{
+            width: '100%',
+            padding: 32,
+            '@media screen and (min-width: 48em)': {
+              width: '50%',
+            },
+            '@media screen and (min-width: 56em)': {
```
vlsi: These kind of selectors prevent consistent look among various window sizes and print view. I'm sure "adaptive" CSS must be opt-in.

packages/themes/src/aspect.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  aspectRatio: 9 / 16,
```
vlsi: Common ratios are 16:9, 4:3, 21:9.
9/16 is understandable, however it is kind of backwards.

PR #340, opened 2019-04-21
fix(docs): export syntax

neighborhood999:


jxnblk:
Thanks! This looks like it might be an issue with the Prettier setup... not sure what's going on here 🤔

PR #354, opened 2019-05-11
fix: stepping logic

aleclarson:


packages/components/src/MDXDeck.js
```diff
@@ -111,7 +111,7 @@ export const MDXDeck = props => {
     navigate(basepath + '/' + nextIndex + search)
     const meta = getMeta(nextIndex)
     setState({
-      step: reverse ? meta.steps || 0 : 0,
+      step: reverse ? (meta.steps || 1) - 1 : 0,
```
aleclarson: When going back to a previous slide, make sure to start on the last step.

packages/components/src/MDXDeck.js
```diff
@@ -130,7 +130,7 @@ export const MDXDeck = props => {
   const next = () => {
     const current = getIndex(props)
     const meta = getMeta(current)
-    if (meta.steps && state.step < meta.steps) {
+    if (meta.steps && state.step < meta.steps - 1) {
```
aleclarson: Keep `state.step` under `meta.steps`

packages/components/src/MDXDeck.js
jxnblk: Is this not currently working as expected? The demo seems to work as intended (see https://mdx-deck.jxnblk.com/9) Might be helpful to have a repro

aleclarson: The current behavior goes to the first step on the previous slide, instead of the last step on the previous slide.

aleclarson: The `meta.steps` is the number of steps, so `state.step` cannot be >= `meta.steps`, and thus we need to do `meta.steps - 1` when `meta.steps` is defined.

jxnblk: Sorry, I'm not following. Again, a reproduction with code would be helpful

aleclarson: The current step begins at 0, which is tracked by `state.step`.

Just like with a JS array, the current step cannot be >= the number of steps.

```js
const steps = [1, 2, 3]
const meta = {steps: 3}
const state = {step: meta.steps || 0}

steps[state.step] // undefined
```

aleclarson: If you have a deck that uses `useSteps` or `<Steps>`, you can try the following:

1. Go to a slide "A" that comes after a slide "B" with 2+ steps
2. Go back to slide "B"
3. Notice how the first step is shown instead of the last
4. Go thru each step in slide "B"
5. Notice how an extra no-op step exists before slide "A"

aleclarson: Also, you can add this snippet to the `<Presenter>` component:
```js
  const { steps } = metadata[index] || {}
  if (props.step >= (steps || 1)) {
    console.warn('Invalid step index: %O / %O', props.step, steps)
  }
```

jxnblk:
Sorry for the delay, but going to close this out without a deployed reproduction to better understand what issue you're trying to solve for. Feel free to open a new issue to discuss

PR #355, opened 2019-05-11
fix(Presenter): next slide preview

aleclarson:
Fix how the next slide is determined by taking steps into account, and pass props into the next slide preview.


jxnblk:
Thanks for working on this and sorry for the delay! Personally, I think the current behavior makes more sense from an interaction design perspective, and since #392 is a fairly large refactor I'm going to close this out. Feel free to open a new issue for discussion if this is something you'd like us to consider

PR #356, opened 2019-05-11
fix: reset notes on every step

aleclarson:
This ensures that the proper notes are shown when navigating backwards. To accomplish this, I've added an `items` prop to the `<Steps>` component. Each item is wrapped in the new `<Step>` component in order to ensure that every `<Notes>` component has access to its associated step. Finally, reset `meta.notes` to an empty string on every slide/step change.

If you have a better solution in mind, let me know! 👍 

It also allows for per-step notes:

```js
import {Steps} from 'mdx-deck'

<Steps
  items={[
    null, // Render nothing on first step
    <Notes>Foo</Notes>,
    <Notes>Bar</Notes>,
  ]}
/>
```

### DIFF
--- a/packages/components/src/Notes.js
+++ b/packages/components/src/Notes.js
@@ -1,15 +1,31 @@
-import React, { useEffect } from 'react'
+import React, { useEffect, useRef } from 'react'
 import { useDeck } from './context'
+import { useStep } from './useStep'
 
 export const Notes = props => {
   const context = useDeck()
+  if (!context || !context.register) {
+    // The slide was rendered as a preview by the <Presenter> component.
+    return false
+  }
+
+  const index = useRef(context.index).current
+  const step = useStep()
   useEffect(() => {
-    if (!context || !context.register) return
-    if (typeof context.index === 'undefined') return
-    context.register(context.index, {
+    // The context changes before this component is unmounted.
+    if (context.index !== index) return
+    // Only render notes for the current step.
+    if (context.step !== step) return
+
+    // Only re-render if the notes changed.
+    const meta = context.metadata[index] || {}
+    if (meta.notes === props.children) return
+
+    // Note: The last <Notes/> per slide always takes precedence.
+    context.register(index, {
       notes: props.children,
     })
-  }, [])
+  })
 
   return false
 }
--- a/packages/components/src/Steps.js
+++ b/packages/components/src/Steps.js
@@ -1,10 +1,30 @@
 import React from 'react'
 import { useDeck } from './context'
 import useSteps from './useSteps'
+import { Step } from './useStep'
 
-export const Steps = props => {
-  const step = useSteps(props.length)
-  return props.render({ step })
+export const Steps = ({ items, render, ...props }) => {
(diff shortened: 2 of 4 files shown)


packages/components/src/Steps.js
```diff
@@ -1,10 +1,30 @@
...
+      {items.map((item, i) => {
+        const visible = props.split ? step === i : step >= i
+        return (
+          visible && (
+            <Step key={i} index={i}>
+              {render
+                ? render(item, i)
+                : typeof item === 'function'
+                ? item(i)
+                : item}
```
aleclarson: When the `items` prop exists, the `render` prop can be used to map the items into elements. Alternatively, you can pass a function in the `items` array prop, and it will be passed its index in the array.

packages/components/src/Notes.js
```diff
@@ -1,15 +1,31 @@
-import React, { useEffect } from 'react'
+import React, { useEffect, useRef } from 'react'
 import { useDeck } from './context'
+import { useStep } from './useStep'
 
 export const Notes = props => {
   const context = useDeck()
+  if (!context || !context.register) {
+    // The slide was rendered as a preview by the <Presenter> component.
+    return false
```
aleclarson: We can break the rule of "no conditional hooks" here, because slides rendered for the `<Presenter>` preview will never be re-render *with* context. They always have **no** context.

packages/components/src/Notes.js
```diff
@@ -1,15 +1,31 @@
...
+  }
+
+  const index = useRef(context.index).current
+  const step = useStep()
```
aleclarson: In my experience, the `context` may not be for the slide that this `<Notes>` component was rendered for. So we need to cache the first `context.index` we get our hands on. For some reason, we can't do the same with `context.step`, as that results in race conditions.

packages/components/src/Notes.js
```diff
@@ -1,15 +1,31 @@
...
-    if (typeof context.index === 'undefined') return
-    context.register(context.index, {
+    // The context changes before this component is unmounted.
+    if (context.index !== index) return
+    // Only render notes for the current step.
+    if (context.step !== step) return
+
+    // Only re-render if the notes changed.
+    const meta = context.metadata[index] || {}
+    if (meta.notes === props.children) return
```
aleclarson: Avoid constantly re-rendering the `<MDXDeck>` component by skipping the `context.register` call when the notes have not changed.

packages/components/src/useStep.js
```diff
@@ -0,0 +1,11 @@
+import { useContext, createContext } from 'react'
+
+const StepContext = createContext(0)
+
+export const useStep = () => useContext(StepContext)
+
+export const Step = props => (
+  <StepContext.Provider value={props.index}>
+    {props.children}
+  </StepContext.Provider>
```
aleclarson: IMO: If we can find a solution that doesn't require this `<Step>` component, we should go with it.

jxnblk:
Thanks for putting this together and sorry for the delay! Going to close this out in favor of the refactoring in #392

PR #357, opened 2019-05-13
fix(loader): syntax errors in dev mode

aleclarson:
The dev server would silently hang when a syntax error was encountered in the @mdx-deck/loader module. You would only see the error after pressing ctrl+c to exit the process manually.

Before: https://streamable.com/7dpju
After: https://streamable.com/0jdz8


jxnblk:
Thanks!

PR #358, opened 2019-05-14
fix: keyboard controls after clicking button or link

aleclarson:
Links and buttons shouldn't prevent keyboard navigation.

### DIFF
--- a/packages/components/src/Keyboard.js
+++ b/packages/components/src/Keyboard.js
@@ -12,7 +12,7 @@ const keys = {
   pgDown: 34,
 }
 
-const inputElements = ['INPUT', 'TEXTAREA', 'A', 'BUTTON']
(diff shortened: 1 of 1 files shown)


jxnblk:
This is to ensure the space bar can click buttons. I think if an element is focused on the page, it’s expected that the keyboard navigation won’t work

aleclarson:
Maybe there should be an option to disable it via the theme or something?

Personally, I don't need spacebar to trigger buttons during my speech.

Diazgin:
This is to ensure the space bar can click buttons. I think if an element is focused on the page,,

jxnblk:
Thanks for working on this, but I think since this could break the behavior in some existing decks, this should be tabled for discussion in an issue and potentially addressed in the next major version

PR #359, opened 2019-05-14
Add useTheme hook

pomber:
Exports the `useTheme` hook so user components can access the theme without adding emotion as a peer dependency.


jxnblk:
Thanks, I'm sure this'll be useful for libraries like Code Surfer :)

Issue #365, opened 2019-05-20
Allow to set Presenter component via theme


codepunkt:
I'm currently working on my own presenter that emphasizes my Notes compared the current and next slides. I've found an easy way to change the Presenter component via theme - which basically requires to merge the themes in `MDXDeck` component instead of the Provider and then, based on the theme, either render the theme-provided Presenter or the default one. The merged could then also be passed to the `Provider` in order to not merge it twice.

It's not a lot of code changes and i'd love to contribute them to not have to maintain my fork. What do you think?

codepunkt:
This would also easily allow things like #46.

Here is a short clip of the presenter that i've built for my presentations - just to give you an idea:
![custom-presenter](https://user-images.githubusercontent.com/603683/58083119-af7e4e00-7bb8-11e9-897b-a3ad7d1543a8.gif)

jxnblk:
Thanks! This is a great idea! I was thinking of supporting this with Gatsby theme's shadowing feature in the next release, but I'll try to take a look at the PR in the next few days

pomber:
I also have my own Presenter, this would be very useful.

codepunkt:
@jxnblk any news on this?

jxnblk:
Sorry, definitely wanna get this in, I just haven't had as much time to focus on mdx-deck recently. Hoping to go through this over the weekend, so hang tight 🙏

PR #366, opened 2019-05-20
feat: themable Presenter

codepunkt:
PR for issue #365

### DIFF
--- a/docs/theming.md
+++ b/docs/theming.md
@@ -125,6 +125,7 @@ The following keys are available for theming:
 - `Slide`: CSS to apply to the wrapping Slide component
 - `components`: object of MDX components to render markdown
 - `Provider`: component for wrapping the entire app
+- `Presenter`: component for wrapping the presenter mode
 - `googleFont`: CSS HREF for adding a Google Font `<link>` tag
 
 ## Advanced Usage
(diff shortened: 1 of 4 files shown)


codepunkt:
Basically:
- `theme` is now assembled in `MDXDeck` and passed as a prop to `Provider`.
- `getWrapper` helper method is moved into `MDXDeck` render method because it needs access to the `theme`
- `Clock` component is exported by `@mdx-deck/components` in order to not have to import it from `@mdx-deck/components/src/Clock` when reusing it in custom `Presenter` components.

alpinegizmo:
+1

Was thinking of doing roughly this myself. The current Presenter doesn't meet my needs, but there probably isn't one best path forward. Give me a clean way to rearrange it myself w/o forking and I'll be happy.

pomber:
hey @jxnblk, any chances to merge this soon? I'm planning to release a new version of code surfer and this would allow me to add a custom presenter too (something like [this](https://twitter.com/pomber/status/1124662753753477120)). 
If you are busy with other projects don't worry, it can wait :raised_hands:

jxnblk:
Sorry for the delay on this! Looks like this works – I think some of the logic in MDXDeck could be refactored into a hook or other utility later, but I think this is good to :shipit:

PR #375, opened 2019-06-06
Cross Browser Sync & Mobile Mode

kuldeepkeshwar:
- Added support for syncing slide state across different browsers (WebSockets)
- Added basic PWA support(manifest & icons)
- Added a new mode for Mobile with Navigation Buttons( sometimes swipe doesn't work for Appear components )
- Prevent vertical scroll

### DIFF
--- a/packages/mdx-deck/README.md
+++ b/packages/mdx-deck/README.md
@@ -305,6 +305,7 @@ See how others have used MDX Deck for their presentations.
 - [Simplify React][simplify-react] by [Kent C. Dodds](https://mobile.twitter.com/kentcdodds)
 - [I Got 99 Problems but GraphQL Ain't One][99-problems] by [Sara Vieira](https://mobile.twitter.com/NikkitaFTW)
 - [Stop de #divFest][stop-div-fest] by [Sara Vieira](https://mobile.twitter.com/NikkitaFTW)
+- [MDX, authors and richer JAMstack content][mdx-talk] by [Josh Dzielak](https://mobile.twitter.com/dzello)
 
 ## Usage Examples
 
@@ -343,3 +344,4 @@ The following examples will open in CodeSandbox.
 [simplify-react]: https://simply-react.netlify.com/#0
 [99-problems]: https://99-problems-graphql-aint-one.now.sh/#0
 [stop-div-fest]: https://stop-div-fest.now.sh/
+[mdx-talk]: https://mdx-talk.developermode.com/


jxnblk:
Sorry for the delay on this and thanks for working on this! Overall looks like a great direction – I'll try to review this over the weekend, so feel free to remind me by pinging a comment here

kuldeepkeshwar:
@jxnblk can you review it?😊

packages/components/src/FluidFontSize.js
```diff
@@ -1,19 +1,25 @@
 // prototype for fluid resizable font size
-import React, { useLayoutEffect, useRef, useState } from 'react'
+import React, { useLayoutEffect, useRef, useState, useEffect } from 'react'
 import ResizeObserver from 'resize-observer-polyfill'
 
 export const FluidFontSize = ({ base = 16, children, className }) => {
   const div = useRef(null)
   const [fontSize, setFontSize] = useState(base)
-
+  const isMountedRed = useRef(false)
```
jxnblk: `s/Red/Ref`?

packages/components/src/MDXDeck.js
```diff
@@ -9,13 +9,15 @@ import defaultTheme from '@mdx-deck/themes/base'
 import Provider from './Provider'
 import Slide from './Slide'
 import Presenter from './Presenter'
+import Mobile from './Mobile'
```
jxnblk: Just starting to dig in here, but I'd kinda prefer to avoid platform-specific code – I'm guessing what you've done would be applicable for iPads as well, right?

packages/components/src/Mobile.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import Zoom from '@mdx-deck/components/src/Zoom'
```
jxnblk: This should use the named import instead of digging into the internal directory structure of the module

packages/components/src/Socket.js
```diff
@@ -0,0 +1,96 @@
+import { useEffect } from 'react'
+
+const STORAGE_INDEX = 'mdx-slide'
+const STORAGE_STEP = 'mdx-step'
```
jxnblk: If these aren't in a shared `constants` module, these probably should be if the `localStorage` logic is still anywhere else

packages/html-plugin/index.js
```diff
@@ -70,12 +70,20 @@ const defaultTemplate = ({
 <html>
   <head>
     <meta charset='utf-8'>
-    <meta name='viewport' content='width=device-width,initial-scale=1'>
+    <meta name='viewport' content='width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=0'/>
```
jxnblk: This is very bad for accessibility, the viewport **needs** to be zoomable

packages/html-plugin/index.js
```diff
@@ -70,12 +70,20 @@ const defaultTemplate = ({
...
     <style>*{box-sizing:border-box}body{margin:0;font-family:system-ui,sans-serif}</style>
     <meta name='generator' content='mdx-deck'>
+    <link rel="manifest" href="/manifest.json">
```
jxnblk: The PWA setup isn't necessary for the touchscreen/mobile UI changes, is it? I'd rather not introduce this here since I have other plans related to this

packages/html-plugin/index.js
```diff
@@ -70,12 +70,20 @@ const defaultTemplate = ({
...
+    <script async="true" src="https://cdn.jsdelivr.net/npm/pwacompat@2.0.9/pwacompat.min.js" integrity="sha384-VcI6S+HIsE80FVM1jgbd6WDFhzKYA0PecD/LcIyMQpT4fMJdijBh0I7Iblaacawc" crossorigin="anonymous"></script>
```
jxnblk: I'm sure this is fairly harmless, but I don't want to load third-party scripts like this. What is this for?

packages/html-plugin/index.js
```diff
@@ -70,12 +70,20 @@ const defaultTemplate = ({
...
     ${head}${links}
   </head>
   <body>
+    <script>
+    function preventScrollBehavior(e) {
+        e.preventDefault();
+    };
+    document.addEventListener("touchmove", preventScrollBehavior, { passive: false });
```
jxnblk: What's the intent here? Is there a reason this isn't in a component?

packages/mdx-deck/lib/static/manifest.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "mdx-deck",
```
jxnblk: This seems like content that should be user-defined – this and the PWA-related code and icons probably should be in a separate issue/PR

packages/components/src/Socket.js
```diff
@@ -0,0 +1,96 @@
...
+function useSocketMessage({ goto, update, index, step }) {
+  useEffect(
+    function() {
+      function handler(event) {
+        const data = JSON.parse(event.data)
+        console.log('Message from server ', data)
+        if (data.type === 'broadcast') {
+          const value = data.payload.value - 0
+          if (data.payload.type === STORAGE_INDEX && value !== index) {
+            localStorage.setItem(STORAGE_INDEX, value)
```
jxnblk: I might be misunderstanding how this works with the existing code, but doesn't the `Storage` component handle this already?

packages/components/src/Mobile.js
jxnblk: Sorry, I misread the line, this should be a local import.

```suggestion
import Zoom from './Zoom'
```

packages/html-plugin/index.js
jxnblk:
```suggestion
    <meta name='viewport' content='width=device-width,initial-scale=1'>
```

packages/components/src/FluidFontSize.js
kuldeepkeshwar: it should be `ref` typo 😩

packages/components/src/Socket.js
kuldeepkeshwar: Yes, `Storage` does the same job. Sometimes(maybe by changing slides at a fast speed) `localstorage` goes out of sync, so first updating the storage & emitting the changes to react.
I couldn't reproduce the exact steps. so commenting these lines as of now.

packages/components/src/Socket.js
kuldeepkeshwar: I will move these to the `constants` module

kuldeepkeshwar:
> Overall looks promising, thanks for putting this together! I hadn't considered using websockets for something like this before.
> 
> For this to land we'd definitely need some docs around how this works for end-users. I also think that would help with the review to make sure I understand how this is intended to work. Unless the PWA-related code is somehow tied to the way the web sockets are meant to work, could you please remove the code related to that from this PR. I will try to open an issue soon that should help explain the rationale related to that.

Oh, my bad. I should have put a detailed context here first 🤦.

(Trying to collect pieces ...)
This PR has 2 parts.
* **Sync deck state across multiple browsers**: Currently deck's state is synced across multiple tabs(within a single browser). The active tab pushes the state in `localstorage` and other tabs have the listener which sync the state. I tried to extend the same idea for multiple browsers using `websockets`. The active tab sends the state to the server(dev) via sockets and server broadcasts the state to other clients and clients have the listener which sync the state.
Related files : [Socket.js](https://github.com/kuldeepkeshwar/mdx-deck/blob/master/packages/components/src/Socket.js), [dev.js](https://github.com/kuldeepkeshwar/mdx-deck/blob/master/packages/mdx-deck/lib/dev.js)

* **PWA**:  When you open the deck on mobile and do `Add to Home Screen`, PWA opens in a `Mobile` mode( similar to other modes we have). It comes with a default `manifest.json` and Icons which you can override by place a  `manifest.json` file at the project root.
Related files : [Mobile.js](https://github.com/kuldeepkeshwar/mdx-deck/blob/master/packages/components/src/Mobile.js), [manifest.json](https://github.com/kuldeepkeshwar/mdx-deck/blob/master/packages/mdx-deck/lib/static/manifest.json)

Motivation: I had a case where I wanted to present the slide & control(navigation) them from mobile, was inclined to convert it to a PWA for a better experience.

PS PWA & Sockets changes are not related. So I will be chopping off PWA related code from this PR

packages/components/src/MDXDeck.js
kuldeepkeshwar: It is just another mode which gets active if you open up with a query `mode=mobile`

packages/html-plugin/index.js
kuldeepkeshwar: `pwacompact` handles the browsers nuisance for PWA. It can be loaded as a script or can be part of the bundle too using `import`

kuldeepkeshwar:
oh no😩, not this one dude 😞(me talking to [pull bot](https://github.com/apps/pull))

kuldeepkeshwar:
@jxnblk can you reopen this pr? [pull bot](https://github.com/apps/pull) closed it  😞.

jxnblk:
Haha! Sorry I've been slow to review PRs in this project... Haven't used pull bot before, but looks like my reopen button is disable?!
Screen Shot 2019-07-02 at 5 19 20 PM

kuldeepkeshwar:
@jxnblk pushed a new commit, please check now. otherwise, I will create a new pr

jxnblk:
Sorry, it's still disabled, but I also don't see any new commits...

kuldeepkeshwar:
@jxnblk opened a new one

PR #386, opened 2019-07-02
Cross Browser Sync & Mobile Mode

kuldeepkeshwar:
dup #375


jxnblk:
Sorry for the delay on this and not to throw a wrench in the work you've already put in, but I'd like to continue with releasing the next major version in #392, which includes a few big changes that I've been wanting to make to the library for a few months now.

I think what you've got here with the socket server is really cool, and I think it'd be a great addition to the new setup in v3. I'll probably be publishing the new version soon, and if you'd like to help get this PR in shape for the changes there, that'd be super cool. But, if you don't have time to work on it, I can probably take what you have and applying it in v3. Thanks again for all the work here, and let me know if you have any questions!

kuldeepkeshwar:
@jxnblk np, I'll update it in the weekend

ejuo:
@jxnblk, do you plan to support this functionality?
I want remote control with mobile phone.
There is [solution](https://dev.to/kevnz/remote-control-an-mdx-deck-presentation-1gmm) from @Kevnz. But it not works for versions >2. Due to this [problem](https://github.com/jxnblk/mdx-deck/issues/604): Provider remounts every render and I end up with a lot of unclosed ws connections. 
There is also a problem that API not give navigate functions.

Issue #383, opened 2019-06-29
MDX Deck v3


jxnblk:
I haven't had as much time to work on MDX Deck lately, so I wanted to document some of the upcoming changes I'd like to introduce in the next major version. A rough prototype was started on the [next](https://github.com/jxnblk/mdx-deck/tree/next/) branch for reference.

The high-level proposed changes are:

- Replace the core MDX Deck CLI with Gatsby
- Use `@mdx-deck/gatsby-theme` as the core part of the app
- Render each slide as static HTML (currently slides are client-side rendered)
- Enable shadowing of the core UI components (Provider, presenter view, etc)
- Allow any relevant Gatsby plugin to work with MDX Deck
- Replace custom theming setup with [Theme UI](https://theme-ui.com)
- Deprecate the use of default exports for layouts in favor of the approach here: #326
- The end-user API for authoring decks should be mostly backwards-compatible

I'm opening this issue to allow for discussion, and to gauge interest if anyone would like to help contribute to working towards the next major version of MDX Deck.

marcysutton:
That all sounds awesome :) I got @mdx-deck/gatsby-theme running today on a new Gatsby site with MDX, and the biggest changes it needed were an upgrade to `gatsby-plugin-mdx` and the `experimentalThemes` API going away. Another thing I noticed about the docs was as someone new to MDX, it would help to list filenames and extensions in code snippets like we do on the Gatsby docs. But I'm super stoked to have a slide deck as part of a site, and it was worth figuring out the setup. Love your idea about rendering static HTML for slides!

jxnblk:
> the biggest changes it needed were an upgrade to gatsby-plugin-mdx and the experimentalThemes API going away.

Thanks, good catch! I'll try to get a PR up to update the docs

> Another thing I noticed about the docs was as someone new to MDX, it would help to list filenames and extensions in code snippets

Are you referring to the MDX code examples in the README.md and other markdown files?

marcysutton:
For the docs thing, it was on the mdx-deck gatsby-theme: https://github.com/jxnblk/mdx-deck/tree/master/packages/gatsby-theme

oorestisime:
Hey excited about this and thanks for laying out the plans for an awesome v3 :)
I am wondering how tightly coupled the themes need to be with mdx-deck. I am currently using grommet for a lot of my projects and have been using it for mdx-deck layouts/themes etc.

[...]

PR #390, opened 2019-07-10
New v3 Gatsby theme

jxnblk:
This is a prototype for the next major version, fully encapsulated as a Gatsby theme. To avoid breaking changes for the time-being, this package is published as `gatsby-theme-mdx-deck`. The components used to render the presentation are all included within this theme, meaning *any* part of the presentation can be customized with component shadowing. Eventually, this package should replace the `@mdx-deck/components` package and also include the `@mdx-deck/layouts` and `@mdx-deck/themes` packages. This also uses Theme UI for theming, which means that themes built for the current version of MDX Deck are not yet fully compatible

Related to #383


jxnblk:
Merging into a working v3 PR

PR #403, opened 2019-07-18
fix basePath issue on single deck site

bkegley:
When adding a custom basePath on a site with only one deck slide transitions move to the index path. `basePath` should be passed to context.

Here's a [repro](https://github.com/bkegley/gatsby-theme-mdx-deck-base-path)


jxnblk:
Awesome, thanks!

PR #417, opened 2019-07-24
Intellij dependency

rummanhasnayeen:
.gitignore update to ignore Intellij IDEA project files.


jxnblk:
You should generally add files like to a global gitignore. Mine includes `.swp` files for example: https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore

PR #418, opened 2019-07-24
Update exporting.md

philmirez:
The exporting documentation was not too straight forward. This is my attempt to help future users out.

I got tripped up on the following:
1. The phrase `MDX dev server` made me feel there was an actual `dev` script, but this was actually referring to the `start` script.
2. The docs assume the MDX server will be running on PORT 8000, but mine defaults to 8080. It took me a second to figure that out. A note about this would save a lot of future users out.

### DIFF
--- a/docs/exporting.md
+++ b/docs/exporting.md
@@ -13,13 +13,23 @@ add a `build` script to your `package.json` file.
 
 ## PDF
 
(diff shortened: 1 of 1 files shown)


jxnblk:
Thanks for this! What version of `mdx-deck` are you using? v2 used port `8080`, but v3 *should* be using `8000`

PR #422, opened 2019-07-28
Use Gatsby `pathPrefix` in theme

jxnblk:


jxnblk:
Tests pass locally, but it looks like Cypress is not working on CI. Going to merge and publish

PR #428, opened 2019-08-05
Fix unmount on every render

pomber:
Slides were unmounting on every render because the `Mode` component was always changing


pomber:
`yarn test` worked locally :man_shrugging:

jxnblk:
Thanks! I *think* the Cypress setup isn't quite right, but this all LGTM!

Issue #440, opened 2019-08-19
Print path is wrong - /print is actually //print


Relequestual:
Using mdx-deck 3.0.9.

I start the server. Presentation opens in browser.
I try to go to `/print` and I get a blank page.
I go to `//print` and I get the presentation in print view.

I look in developer tools and see page-data.json defined `allSitePages` nodes with paths of `/`and `//print`.

I discovered this when my pdf export resulted in a blank PDF.

Running `npx website-pdf http://localhost:8000//print -o deck.pdf` is now what's required to generate my PDF. I would guess double slash print is not what is expected, right?

iamchathu:
This also downloads chromium every time of 108MB.

PR #454, opened 2019-08-27
`/print` requires two slashes to work

keepkeyjon:


pvinis:
Here is the issue about it: https://github.com/jxnblk/mdx-deck/issues/440

jxnblk:
Thanks, but the route for `/print` should be fixed, not a change in the docs

PR #457, opened 2019-08-29
Add plugin to import code blocks from files

pomber:
Added [gatsby-remark-import-code](https://github.com/pomber/gatsby-remark-import-code) to the gatsby-plugin-mdx config.

This is particularly useful for people using Code Surfer, but it can also be used with any "vanilla" code block.  

It transforms code blocks from:

````md
```js file=./hello-world.js
```
````

into:

````md
```js
function helloWorld() {
  return "hello world";
}
```
````


jxnblk:
Thanks! Seems like a good addition

PR #465, opened 2019-09-03
Fix slug generation in windows

joshrasmussen34:
When building on a windows machine, slugs maintain a windows path
format (i.e., slides were located at http://localhost:8000/\hello)

Forced npm path module to use posix formats to fix this.


jxnblk:
Thanks!

PR #473, opened 2019-09-10
Use correct path for print (fixes #440)

### LINKED ISSUE #440: Print path is wrong - /print is actually //print

robinpokorny:
When the `basePath` is exactly `"/"` then the resulting path for printing is `//print`. See #440 

This PR makes sure there is no double slash reusing logic needed for matching unknown paths.

### DIFF
--- a/packages/gatsby-theme/gatsby-node.js
+++ b/packages/gatsby-theme/gatsby-node.js
@@ -116,7 +116,7 @@ exports.createPages = async ({ graphql, actions, reporter, pathPrefix }) => {
       },
     })
     createPage({
-      path: basePath + '/print',
+      path: base + '/print',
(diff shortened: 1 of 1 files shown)


jxnblk:
Awesome, thanks!

Issue #474, opened 2019-09-10
mdx-deck 3.0.10, node v11.5.0 - Hot updates only render if first slide is modified


jonsherrard:
```
Test

---

Test 2
```

->

```
Test

---

Test 2!
```

= No Update

```
Test!

---

Test 2!
```

= Both updates appear. 

Really strange one.

## Update

To fix your mdx-deck v3 project temporarily, run the following commands in your presentation project:

```
rm -rf node_modules
npm install gatsby@2.14.7 -D --save-exact
npm install
```

jxnblk:
Can you create a reproduction? This *should* be working

jonsherrard:
Sure here you go: https://github.com/jonsherrard/mdx-repro

jxnblk:
Thanks! I'm seeing this too locally – very weird... will need to dig into this a bit later

jonsherrard:
OK that's good to know - I'd have a prod around but it's changed so much since moving to Gatsby!

It also happens if you have multiple slides, not just two. Only the first slide registers as an update. Where does MDX-Deck split the mdx file on the `thematicBreak` node? I could have a nosey around too...

jxnblk:
FWIW most of the relevant code is probably in the `gatsby-theme-mdx-deck` package now, this is the wrapper component for splitting slides: https://github.com/jxnblk/mdx-deck/blob/master/packages/gatsby-theme/src/split-slides.js which is used here https://github.com/jxnblk/mdx-deck/blob/f9bc2456ff233157c74d5f9e0029eff25417f11b/packages/gatsby-theme/src/templates/deck.js

jonsherrard:
Awesome thanks!

patlux:
Thanks. I already thought I'm the only one with this problem. I just downgraded to v2.5.1 as temporary solution till the bug is fixed in v3.

jxnblk:
I believe this is fixed in #493 (v3.0.13), but please reopen if the issue persists

jonsherrard:
Thank you @jxblnk! super appreciative of mdx-deck. Owe you one! Sorry I couldn’t figure it out 😓

PR #477, opened 2019-09-12
Pin underlying Gatsby to patch releases of v2.14

jonsherrard:
Fixes: https://github.com/jxnblk/mdx-deck/issues/474

There's a fairly major update to Gatsby from 2.14, and 2.15. This will keep `mdx-deck` working for fresh install in the meantime


jxnblk:
Thanks! I think this should be fixed in #493 – generally I don't want to restrict versions of Gatsby like this, so if the issue still persists, it should be fixed to work with the latest version of Gatsby

PR #502, opened 2019-10-01
Use opt+f to go fullscreen

ravanscafi:
I was having trouble with going fullscreen on firefox (the navbar would not go away on fullscreen, not sure why), so I created this logic for a presentation myself.
I think it's a cool feature and could help others, so I created this PR.

I used a simple "polyfill" just in case, but I can change anything as desired, add more docs, etc :)

### DIFF
--- a/docs/presenting.md
+++ b/docs/presenting.md
@@ -6,6 +6,7 @@
 3. Move the new tab to its own window in the screen or projector that the audience sees
 4. Control the presentation from the original window
 5. Be sure to hide your mouse cursor so that it's not visible to the audience
+6. You can press `Opt + F` on the new tab to go fullscreen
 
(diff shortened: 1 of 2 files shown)


nkuehn:
Great addition that I wanted to do myself, too to be able to get "real" fullscreen presentations in all situations. 

I have tested it and it also works when two windows are open in presenter mode.   If you want to polish, you could also catch the `msRequestFullscreen` vendor prefix and catch the case when none is found at all. 

I can't merge though (not the maintainer).

aszx87410:
@jxnblk I was wondering that is there any plan to merge this PR? 
This feature is very helpful and it's a very straightforward way to make it fullscreen. It would be great if this can be merged.

PR #503, opened 2019-10-02
Added hot pink theme using emotion-theming

complex-coder:
Used emotion-theming library and added a hot pink theme take a look and let me know.


jxnblk:
Thanks! But you shouldn't need to use the `ThemeProvider` directly, since this library uses Theme UI, see https://github.com/jxnblk/mdx-deck/blob/master/docs/theming.md

PR #513, opened 2019-10-12
Improve handling of eject.

### LINKED ISSUE #511: eject not working
Using `mdx-deck eject` is currently not working as it should.

This is because the logic in [initit](https://www.npmjs.com/package/initit) does not match what the `eject` command expects. So either we send a PR to `initit` changing the logic (changing how it handles absolute paths) or we adapt `eject` to the actual logic/result/structure of `initit`.

Adapting `eject` to `initit` would basically require `eject` to have intricate knowledge of the logic inside `initit` and that doesn't feel right to me. I would rather try and get `initit` to alter the logic it uses and support absolute paths in the way I assume that `eject` expects.

## `initit` logic details

The logic inside `initit` is that it takes a `name` argument which it uses to create a folder in the current working directory. This makes sense if used as intended, you pass a string containing only the name of the directory you want to initialize in.

But `eject` doesn't pass a name, it passes the `cwd`, meaning an absolute path. Since `initit` expects a name, not an absolute path, it tries to resolve an absolute path based on the given name. This is where it gets tricked.

 Given a "name":
```
cwd = /tmp/my-cwd
name = my-name
resolves: /tmp/my-cwd/my-name
creates: /tmp/my-cwd/my-name
```

Given an absolute path:
```
cwd = /tmp/my-cwd
name = /tmp/my-cwd
resolves = /tmp/my-cwd
creates = /tmp/my-cwd/my-cwd
```
As you see `initit` creates the given folder name based on the `basename` of the resolved path.

What I (and I assume also `eject`) expected was:
```
cwd = /tmp/my-cwd
name = /tmp/my-cwd
resolves = /tmp/my-cwd
creates (reuses really) = /tmp/my-cwd
```

I assume that `eject` expected `initit` to reuse the current working directory. If it relied on the current implementation described above, it nees to adjust its code to reflect the resulting directory structure from `initit`. Specifically this line `fs.moveSync(filename, path.join(cwd, 'decks'))` needs to change to something like `path.join(cwd, path.basename(cwd), `decks/${path.basename(filename)}`)` to match up with the resulting structure from the current `initit`. 

I'm willing to send a PR both ways, would just like some guidance in what would be the accepted approach.

michaellopez:
As discussed in [#511](https://github.com/jxnblk/mdx-deck/issues/511#issuecomment-541092796)

This is my initial stab at improving `eject`.

It removes `initit` as a dependency and instead manipulates the current working directory to look like [the starter](https://github.com/jxnblk/mdx-deck/tree/master/packages/starter) and uses `execa`, that was already a dependency, to run `npm install`.

Was it something like this you recommended @jxnblk ?

Fixes #511

### DIFF
--- a/packages/mdx-deck/cli.js
+++ b/packages/mdx-deck/cli.js
@@ -96,10 +96,23 @@ switch (cmd) {
     })
     break
   case 'eject':
-    log('ejecting Gatsby site')
+    log('ejecting from mdx-deck into Gatsby site')
+    log('this can take a few minutes...')
     require('./eject')({
       cwd: process.cwd(),
-      filename: path.resolve(filename),
+      file: path.resolve(filename)
+    }).then(() => {
+      log('ejected successfully')
+      log(`you can now run the following commands:`)
+      console.log('');
+      console.log(`  ${chalk.cyan('npm start')}`)
+      console.log(`    starts the Gatsby development server`)
+      console.log('')
+      console.log(`  ${chalk.cyan('npm run build')}`)
+      console.log(`    creates a build of the Gatsby project`)
+      console.log('')
+      console.log(`  ${chalk.cyan('npm run clean')}`)
+      console.log(`    wipes the locally built gatby assets and cache`)
     }).catch(err => {
       log.error(err)
     })
--- a/packages/mdx-deck/eject.js
+++ b/packages/mdx-deck/eject.js
@@ -1,13 +1,55 @@
 const path = require('path')
 const fs = require('fs-extra')
-const initit = require('initit')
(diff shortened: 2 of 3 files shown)


michaellopez:
Also I didn't really find any tests, if I missed them please point me in the right direction and I will have a look 😄

jxnblk:
Thanks! This looks great! I don't think there were any e2e tests for this before, so no worries about testing. I think if this works with monkey-testing, that should be fine for functionality like this and since the code is fairly straightforward.

I'll try to pull this down in a bit and try it out locally before merging

PR #527, opened 2019-10-24
Update docs & examples to use v3 package imports

qjcg:
This fixes the docs pages and examples that still used the incorrect v2 package names (`@mdx-deck/themes`, `@mdx-deck/components`).

### DIFF
--- a/docs/themes.md
+++ b/docs/themes.md
@@ -1,9 +1,5 @@
 # Themes
 
-```sh
-npm i mdx-deck/themes
-```
-
 ![Default theme](images/default.png)
 
 ---
--- a/docs/themes.md
+++ b/docs/themes.md
@@ -1,7 +1,7 @@
 # Themes
 
 ```sh
-npm i @mdx-deck/themes
+npm i mdx-deck/themes
 ```
 
 ![Default theme](images/default.png)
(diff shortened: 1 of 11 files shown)


docs/themes.md
```diff
@@ -1,7 +1,7 @@
 # Themes
 
 ```sh
-npm i @mdx-deck/themes
+npm i mdx-deck/themes
```
jxnblk: Thanks for taking a stab at updating the docs! This isn't a valid package name, so I think this should be removed

qjcg: Indeed, that's my mistake. Now fixed with my latest commit.

docs/themes.md
```diff
@@ -1,7 +1,7 @@
 # Themes
 
 ```sh
-npm i mdx-deck/themes
+npm i mdx-deck
```
jxnblk: I meant, I don't think it makes sense to repeat the install instructions here since it's now the same package. I think this entire code fence can be removed

qjcg: Aha, I missed that! The entire code-fenced block's now been removed in my latest commit.

jxnblk:
Thanks! 🎉

PR #550, opened 2019-11-08
Suggest exporting via "Print to PDF"

borekb:
All modern operating systems include this option so I think it's easier than to fetching an npm package that downloads its own Chromium.

### DIFF
--- a/docs/exporting.md
+++ b/docs/exporting.md
@@ -13,13 +13,15 @@ add a `build` script to your `package.json` file.
 
 ## PDF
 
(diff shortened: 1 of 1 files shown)


jxnblk:
You can navigate to `/print` to print as PDF from your browser. The npm package uses puppeteer to do this in Node.js

borekb:
Yep, I know now but `docs/exporting.pdf` don't mention that and it took me some time to figure that out. So I thought a mention in docs would help someone in the future.

Issue #374, opened 2019-06-06
can't print slide numbers


alpinegizmo:
It doesn't seem to be possible to have slide numbers printed out. 

I've implemented a custom Provider that puts props.index into a fixed position on each slide. This works great in normal mode. However, in print mode, this prints as a 0 for every slide.

I tried to rework the page numbers so they are position: relative, but that became a mess and I gave up.

jxnblk:
Sorry for the delay! With the way the components are set up, I don't think it's possible to do this in the print view

PR #626, opened 2020-01-14
Allows the usage of slide index in print mode

foyarash:
Refers to #374 

Allows to retrieve the slides index using the `useDeck` hook on print mode. Previously the value was always 0, because on the `Print` component, the slide index is not passed to the `Slide` component.

### DIFF
--- a/packages/gatsby-theme/src/components/deck.js
+++ b/packages/gatsby-theme/src/components/deck.js
@@ -37,7 +37,7 @@ const Print = ({ slides }) => {
   return (
     <Context.Provider value={context}>
       {slides.map((slide, i) => (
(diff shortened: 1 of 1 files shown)


alpinegizmo:
Even better would be to do this, which makes it possible to print the notes as well.

```js
const Print = ({ slides, notes }) => {
  const outer = useDeck()
  const context = {
    ...outer,
    mode: modes.print,
  }

  return (
    <Context.Provider value={context}>
      {slides.map((slide, i) => (
        <Slide key={i} slide={slide} index={i} notes={notes} preview />
      ))}
    </Context.Provider>
  )
}
```

maxammann:
Would be very cool to see this merged :)

PR #628, opened 2020-01-26
Using constant for normal mode on step 0

pedroapfilho:
I came across this issue when I was developing a theme, and I was checking if the user was in "normal" mode, I saw that it starts with "normal", but if you hit `escape`, it becomes "NORMAL", capitalised.

I checked the code and found out that the constants are not used on the `app.js`,  so I fixed it.


jxnblk:
Awesome, thanks!

PR #651, opened 2020-02-18
[WIP] Improve presenter mode

### LINKED ISSUE #596: Scrollable presenter notes
I have a slide with lots of presenter notes. I know that's not great, however it would be awesome if they didn't overflow the page but instead scrolled:

I think it should be a matter of adding a `max-height` and `overflow-y: scroll` to the `Notes` component.

zyhou:
Hello,

I found notes very small, it's not a good pratice to put to much note but you know..

I add multiple presenter mode:
- Normal
- Tall
- Wide
- Only notes

presenter-mode

## Todo

- [ ] Improve design
- [ ] Add test

Not the solution from this issue but can resolve #596

### DIFF
--- a/packages/gatsby-theme/src/components/app.js
+++ b/packages/gatsby-theme/src/components/app.js
@@ -1,7 +1,7 @@
 import React, { useReducer } from 'react'
 import merge from 'lodash.merge'
 import Context from '../context'
-import { modes } from '../constants'
+import { modes, presenterModes } from '../constants'
 
 const reducer = (state, next) =>
   typeof next === 'function'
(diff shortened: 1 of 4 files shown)


zyhou:
@jxnblk My PR is ready for review but I leave in work in progress. I change the gatsby-theme packages but this is not use for mdx-deck package. I need to update the gatsby-plugin package?  [Presenter component](https://github.com/jxnblk/mdx-deck/blob/master/packages/gatsby-plugin/src/container.js#L49)

I'am not sure why we need to duplicate functionnality between these package.

Issue #654, opened 2020-02-22
Error with Version 4.1.0


forjyoung:
nfo Deleting .cache, public
info Successfully deleted directories

 ERROR #10226  CONFIG

Couldn't find the "gatsby-source-filesystem" plugin declared in "/Users/bryce/docs/code-surfer/mdx-deck/node_modules/mdx-deck/gatsby-config.js".

Tried looking for a local plugin in /Users/bryce/docs/code-surfer/mdx-deck/node_modules/mdx-deck/plugins/gatsby-source-filesystem.

Tried looking for an installed package in the following paths:
 - /Users/bryce/docs/code-surfer/mdx-deck/node_modules/gatsby/dist/bootstrap/load-themes/node_modules/gatsby-source-filesystem
 - /Users/bryce/docs/code-surfer/mdx-deck/node_modules/gatsby/dist/bootstrap/node_modules/gatsby-source-filesystem
 - /Users/bryce/docs/code-surfer/mdx-deck/node_modules/gatsby/dist/node_modules/gatsby-source-filesystem
 - /Users/bryce/docs/code-surfer/mdx-deck/node_modules/gatsby/node_modules/gatsby-source-filesystem
 - /Users/bryce/docs/code-surfer/mdx-deck/node_modules/gatsby-source-filesystem
 - /Users/bryce/docs/code-surfer/node_modules/gatsby-source-filesystem
 - /Users/bryce/docs/node_modules/gatsby-source-filesystem
 - /Users/bryce/node_modules/gatsby-source-filesystem
 - /Users/node_modules/gatsby-source-filesystem
 - /node_modules/gatsby-source-filesystem

not finished open and validate gatsby-configs - 0.055s
Error: Command failed with exit code 1: gatsby develop --host localhost --port 8000 --open
    at makeError (/Users/bryce/docs/code-surfer/mdx-deck/node_modules/execa/lib/error.js:58:11)
    at handlePromise (/Users/bryce/docs/code-surfer/mdx-deck/node_modules/execa/index.js:114:26)
    at processTicksAndRejections (internal/process/task_queues.js:89:5)

gopinav:
Can confirm this is happening in 4.1.0. Wasn't happening previously. I had to manually run `yarn add -D gatsby-source-filesystem` to make it work for now.

sebald:
@gopinav Hm, when I add it, I end up in an endless watch/build loop 😭

estaub:
[...]

PR #655, opened 2020-02-23
Add missing dep.

millette:
Fix for #654

Add missing gatsby-source-filesystem dep.


jxnblk:
Thanks!

spawnia:
Thanks for the great package and this fix. A bugfix release with this would be great, just ran into this issue.

PR #656, opened 2020-02-24
Add "intro to webdev" example deck.

HoverBaum:
Adds an example deck to the readme that illustrates use of some custom components.


jxnblk:
Thanks!

PR #667, opened 2020-03-03
Fix markdown syntax for example deck.

HoverBaum:
The last pull request introduce a wrong markdown syntax leading to a weird output of link-text and link instead of a text that is a link.

This Pull Request fixes that.

Sorry for the inconvinience.


jxnblk:
Ha! Thanks!

Issue #425, opened 2019-07-31
Look into properly supporting `pathPrefix` in Gatsby


jxnblk:
While it halfway works, setting `pathPrefix` in a Gatsby config causes some odd behavior in the theme. This might be due to the use of `matchPath` + how the base path is passed to React Router

liltechnomancer:
Is this for the basePath options passed through Gatsby Config? If so I think I am running into what you are talking about. It seems to just continue adding that base path on the end of the url. 

If I set it to:
```
    options: {
        basePath: "/slides",
      },
```

and visit localhost:8000/slides/intro and then click the -> arrow the URL becomes "http://localhost:8000/slides/slides/intro/1"

if I start on http://localhost:8000/slides/intro/1 and hit the <- arrow the URL becomes 
"http://localhost:8000/slides/intro/slides/intro/0"

I could maybe look into fixing this but im not sure where to start.

rohit-gohri:
> Is this for the basePath options passed through Gatsby Config? If so I think I am running into what you are talking about. It seems to just continue adding that base path on the end of the url.
> 
> If I set it to:
> 
> ```
>     options: {
>         basePath: "/slides",
>       },
> ```
> 
> and visit localhost:8000/slides/intro and then click the -> arrow the URL becomes "http://localhost:8000/slides/slides/intro/1"
> 
> if I start on http://localhost:8000/slides/intro/1 and hit the <- arrow the URL becomes
> "http://localhost:8000/slides/intro/slides/intro/0"
> 
> I could maybe look into fixing this but im not sure where to start.

Did you ever fix this? Not sure if this is fixed in latest version as I can't use v4 because it breaks with code-surfer (#485).

PR #674, opened 2020-03-21
Use absolute paths for navigation

rohit-gohri:
This fixes the problem with basePath mentioned in https://github.com/jxnblk/mdx-deck/issues/425#issuecomment-546048204
Not sure if the actual issue was also about that?


rohit-gohri:
Just realized that this breaks Presentation Mode. Presentation Mode turns off on every page change.

PR #681, opened 2020-03-31
Ignore shortcuts when elements are focused

dojr:
Was trying to add inputs to slide deck and keyboard shortcuts kept moving to the next slide. This code is also in gatsby-theme so I just moved it to also be in gatsby-plugin.

I added this code and tested it locally and it works 

Let me know what you think and if there any other additional things I need to do.

### DIFF
--- a/packages/gatsby-plugin/src/keyboard.js
+++ b/packages/gatsby-plugin/src/keyboard.js
@@ -16,6 +16,8 @@ const keys = {
   pageDown: 34,
 }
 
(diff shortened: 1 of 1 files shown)


scmilee:
This is sick!!

donaldjr:
@jxnblk any chance I could get a review on this? I realize I could probably use the gatsby plugin instead since this logic is also in there but I didn't want to bring in gatsby to my project unless thats the recommended usage for mdx deck.

PR #700, opened 2020-04-11
fix: Stop slides keyboard navigation if user might be typing

okonet:
This stops keyboard navigation if user has input / textarea focused and might be typing.

Basically a copy-paste from https://github.com/jxnblk/mdx-deck/blob/d3cea6b44a1f252432313ac4f6a5a7cc4f5844a6/packages/gatsby-theme/src/hooks/use-keyboard.js#L41-L43 which made me wondering why there are 2 versions of hooks etc.

PS: I'm using stand-alone version of mdx-deck it has this bug still.

### DIFF
--- a/packages/gatsby-plugin/src/keyboard.js
+++ b/packages/gatsby-plugin/src/keyboard.js
@@ -16,6 +16,8 @@ const keys = {
   pageDown: 34,
 }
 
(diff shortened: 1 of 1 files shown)


dojr:
#681

PR #729, opened 2020-06-11
Docs: update themes docs to work and match contemporary package scopes

cspeterson:


wrongbaud:
Thanks for this, I was struggling to figure this out.

PR #739, opened 2020-07-24
Fix #738 - CodeSurfer animations briefly flashing

### LINKED ISSUE #738: CodeSurfer animations briefly flash all at once when entering a slide
Hi!

First of all, thanks for the great tool! :) 

I found an issue with the latest (4.1.1) version of mdx-deck, and I'm also submitting a PR to fix it.

If you have a slide with a `<Steps>` component before a slide with a (multi-step) `<CodeSurfer>` component, as soon as the slide with the `<CodeSurfer>` component is loaded, you'll see all CodeSurfer animations played back in reverse, very quickly.

To see this happen, install CodeSurfer in the demo project, and add these two slides to `demo.mdx`:

    ---
    
    # Steps
    
    <Steps>
    
    - these
    - are
    - some
    - steps
    
    </Steps>
    
    ---
    
    # CodeSurfer
    
    <CodeSurfer>
    
    ```js
    console.log("hey!");
    ```
    
    ```js
    console.log("hey!");
    console.log("I'm a test");
    ```
    
    ```js
    console.log("hey!");
    console.log("I'm a test");
    console.log("for CodeSurfer");
    ```
    
    ```js
    console.log("See all those animations at the beginning?");
    ```
    
    ```js
    console.log("they shouldn't show up");
    ```
    
    </CodeSurfer>
    
    ---

I'm creating a new PR to fix this behaviour. It worked for me, but I don't know if it's the best way to fix it :)

micheleb:
The issue I reported happens because the current page index is reset before the current step is, so CodeSurfer renders the diff between its last step and the first (hence, it displays them backwards, very quickly).

The only change I had to make was inverting the two calls at lines 92-93. All the other changes were added by prettier via the pre-commit hook.

### DIFF
--- a/packages/gatsby-plugin/src/deck.js
+++ b/packages/gatsby-plugin/src/deck.js
@@ -23,9 +23,8 @@ export default props => {
   const slide = slides[index]
 
   const [mode, setMode] = React.useState(modes.default)
-  const toggleMode = next => setMode(current =>
-    current === next ? modes.default : next
-  )
+  const toggleMode = next =>
+    setMode(current => (current === next ? modes.default : next))
 
   const [step, setStep] = React.useState(0)
   const [steps, setSteps] = React.useState(0)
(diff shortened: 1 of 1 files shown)


erdemu:
Accurate description and solution. Would recommend attention to this. Maybe removing all the ide changes and only doing the reordering could be better option before a merge. This way it could be easier in the future to follow code evolution. (If you blame ideally you shouldn't see anything unrelated to the fix commit itself in the history)

PR #743, opened 2020-08-15
update theme documentation

M0nica:
I updated the theme documentation as the import statements were outdated and how to set the them was a bit unclear. 

Relates to: https://github.com/jxnblk/mdx-deck/issues/741 and https://github.com/jxnblk/mdx-deck/issues/699

PR #744, opened 2020-08-19
Fix/improve themes examples

haraldreingruber:
It took me a bit to figure out how to change then theme, as I was looking first at the examples and missed that the `theme` const needs to be exported as well.

I kinda prefer importing the whole `themes` object from `'mdx-deck'`, but I also like the previous style of importing only the theme which is actually used. If that's preferred, I guess the package name changed recently and we would need to do something like:

```js
import { big } from '@mdx-deck/themes'
export const theme = big
```

### DIFF
--- a/docs/themes.md
+++ b/docs/themes.md
@@ -7,7 +7,8 @@
 ![Big theme](images/big.png)
 
 ```js
-import { big } from 'mdx-deck/themes'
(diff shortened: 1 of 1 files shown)


haraldreingruber:
Just saw that there is already a PR for this: #743 
Thanks, @M0nica

Issue #759, opened 2020-10-18
Readme Clarity


kristopolous:
This is a very simple fix.

At first blush, I had no idea where the ES6 style code in the readme was supposed to go (a separate file, maybe some specially named place? A configuration file?) and I was extremely confused as to where to place it as it was not specified.

I've resolved this with a small bit of text. PR coming.

PR #760, opened 2020-10-18
Some readme clarity

kristopolous:
A very small fix that clarifies instructions in the readme.


kristopolous:
Linking it over Fixes #759

